Issue #46, opened 2020-02-10
Add ability to notify bootstrap consumer of config update


michaelestrin:
Notify go-mod-bootstrap caller/consumer when configuration provider is used and a remote configuration change is made.

Required for AppFunc.

Issue #44, opened 2020-02-08
Implement commandLine.NewDefaultCommonFlagsWithoutUsage()


michaelestrin:
https://github.com/edgexfoundry/edgex-go/pull/2317#discussion_r376609586

Will require follow-on PR for edgex-go.

PR #47, opened 2020-02-10
Add ability to notify bootstrap consumer of config update

michaelestrin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/edgexfoundry/go-mod-bootstrap/issues/46

## What is the new behavior?

When provided, a stream passed to RunAndReturnWaitGroup() is called when a configuration provider is used and a remote configuration change is made.

Dependent upon https://github.com/edgexfoundry/go-mod-bootstrap/issues/44

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -31,8 +31,15 @@ import (
 
 const writableKey = "/Writable"
 
+// UpdatedStream defines the stream type that is notified by ListenForChanges when a configuration update is received.
+type UpdatedStream chan<- struct{}
+
 // createClient creates and returns a configuration.Client instance.
-func createClient(serviceKey string, providerConfig configTypes.ServiceConfig, configStem string) (configuration.Client, error) {
+func createClient(
+	serviceKey string,
+	providerConfig configTypes.ServiceConfig,
+	configStem string) (configuration.Client, error) {
+
 	providerConfig.BasePath = configStem + serviceKey
 	return configuration.NewConfigurationClient(providerConfig)
 }
@@ -95,7 +102,8 @@ func ListenForChanges(
 	wg *sync.WaitGroup,
 	config interfaces.Configuration,
 	lc logger.LoggingClient,
-	configClient configuration.Client) {
+	configClient configuration.Client,
+	stream UpdatedStream) {
 
 	wg.Add(1)
 	go func() {
@@ -129,6 +137,10 @@ func ListenForChanges(
 
 				lc.Info("Writeable configuration has been updated from the Configuration Provider")
 				_ = lc.SetLogLevel(config.GetLogLevel())
+
+				if stream != nil {
+					stream <- struct{}{}
+				}
 			}
 		}
 	}()
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -81,18 +81,18 @@ func RunAndReturnWaitGroup(
 	serviceKey string,
 	configStem string,
 	serviceConfig interfaces.Configuration,
+	configUpdatedStream config.UpdatedStream,
 	startupTimer startup.Timer,
 	dic *di.Container,
 	handlers []interfaces.BootstrapHandler) (*sync.WaitGroup, bool) {
 
-	lc := logging.FactoryToStdout(serviceKey)
 	var err error
-	var configClient configuration.Client
-	var registryClient registry.Client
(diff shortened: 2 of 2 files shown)


bootstrap/config/config.go
```diff
@@ -25,15 +25,21 @@ import (
...
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
 )
 
-const (
-	WritableKey = "/Writable"
-)
+const writableKey = "/Writable"
+
+// UpdatedStream defines the stream type that is notified by ListenForChanges when a configuration update is received.
+type UpdatedStream chan<- struct{}
```
lenny-intel: Need to send copy of previous Writable before updated, so I think needs to be `interface{}` rather than `struct{}`.

bootstrap/config/config.go
```diff
@@ -129,7 +136,11 @@ func ListenForChanges(
 				}
 
 				lc.Info("Writeable configuration has been updated from the Configuration Provider")
-				lc.SetLogLevel(config.GetLogLevel())
+				_ = lc.SetLogLevel(config.GetLogLevel())
+
+				if stream != nil {
+					stream <- struct{}{}
```
lenny-intel: Send copy of config prior to updating on line 133. This is because we have to detect what changed. See https://github.com/edgexfoundry/app-functions-sdk-go/blob/master/appsdk/sdk.go#L719

michaelestrin:
Rebased.  Moved from draft status.

bootstrap/config/config.go
michaelestrin: There's no clean way to do this.  

If I move this to where you want it, there's the potential for the follow-on call to config.UpdateWritableFromRaw() to fail.  Further, you'd be receiving a copy of the current (as yet unmodified) configuration on the stream; there's a potential race condition hidden in there.

If I don't move this, I'd have to copy the underlying structure before calling config.UpdateWritableFromRaw() and subsequently putting the copied struct onto the stream.

IMO, this is a caller's concern; delta detection should be handled there.  You're already likely going to create a goroutine to handle the inbound change notifications.  You can keep your own copy of the configuration struct there to diff with the latest (the same reference that ListenForChanges() is updating).

What do you think @lenny-intel?

lenny-intel: @michaelestrin Yes, delta detection is caller's concern as long as it has the data to do it. We could grab the initial copy after RunAndReturnWaitGroup() returns. But this will only work for SDKs. If one of the core services decided it need to do something similar, it would have opportunity to grab the initial copy since Run() doesn't return.  

I still think the better solution is to have the copy sent back on the channel.

The race condition can be avoided by simply grabbing a copy before calling config.UpdateWritableFromRaw() and then sending it over the channel where you are currently writing to the channel.  i.e. no moving required. The caller would expect this to be the previous values knowing the current values have already been applied by config.UpdateWritableFromRaw().

Question is then how to you get a copy? I suggest new interface config.CopyOfWritable() which returns an interface{}.

Thoughts?

michaelestrin: > If one of the core services decided it need to do something similar, it would have opportunity to grab the initial copy since Run() doesn't return.

This is an fence-post edge case and easily solved.  As in your use case, a core service would spin a separate go routine in its service-specific bootstrap handler to handle the updates.  By the time the service-specific bootstrap handler is executed, configuration has already been populated and is available to copy.

> I still think the better solution is to have the copy sent back on the channel.

I don't agree.  You're proposing to add a `CopyOfWritable()`.  This increases complexity and expands the surface area of the contract -- both undesirable.  Further, it would require implementation for concrete instances where it would never be called.

It sounds like you are able to capture your own copy of your service's configuration as part of start up and use that copy to determine what specific properties have changed when a notification is received from the stream.

I believe core services can do the same (notwithstanding that Run() doesn't return until service exit).

Why not take this path?

lenny-intel: Yep, I think the bootstrap handler will work as long as it has access to the initial config, which we should be able to make happen... I wasn't thinking that the new go routine would be in a bootstrap handler.

michaelestrin: Doesn't have to be for you.  Does for core. ;)

PR #58, opened 2020-03-20
Feature/fix issue 2200

tingyuz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

go-mod-boostrap only exposes interface of retrieving database credentials. 

Issue Number:
https://github.com/edgexfoundry/edgex-go/issues/2200

## What is the new behavior?
Adding new interface to expose a method of retrieving SSL cert key pair.

- [ ] Yes
- [X] No

## Are there any specific instructions or things that should be known prior to reviewing?
The new interface CertificateProvider and the method of GetCertificateKeyPair() will be helpful for security-proxy-setup as well as other core microservices in the future.
1. The security-proxy-setup already adopted go-mod-bootstrap module. The new interface will not break its existing behaviors.
2. When the environment variable EDGEX_SECURITY_SECRET_STORE is set up as false which means security for EdgeX is off, the user will use a docker-compose file that doesn't include any security component.
3. There is a plan to add certificate key pair for core microservices ( such as data, metadata, commend etc.) to enable Https in the future. This interface and method can be used for such security enhancement in the future. 

## Other information

### DIFF
--- a/bootstrap/container/certificate.go
+++ b/bootstrap/container/certificate.go
@@ -1,0 +1,29 @@
+/********************************************************************************
+ *  Copyright 2020 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/interfaces"
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
+)
+
+// CertificateProviderName contains the name of the interfaces.CertificateProvider implementation in the DIC.
+var CertificateProviderName = di.TypeInstanceToName((*interfaces.CertificateProvider)(nil))
+
+// CertificateProviderFrom helper function queries the DIC and returns the interfaces.CertificateProviderName
+// implementation.
+func CertificateProviderFrom(get di.Get) interfaces.CertificateProvider {
+	return get(CertificateProviderName).(interfaces.CertificateProvider)
+}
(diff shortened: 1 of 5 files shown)


bnevis-i:
See https://github.com/edgexfoundry/edgex-go/issues/2200

bnevis-i:
- This PR has multiple commits and should be reduced to a single commit.
- Commit description should follow conventional commits naming style.
- No unit test for the newly added code
- Not sure if need to abbreviate Certificate to Cert and PrivateKey to Key...

Otherwise straightforward and LGTM with the general approach.  Did not realize that bootstrap did not provide a way to get to the secretstoreclient directly.

michaelestrin:
> Did not realize that bootstrap did not provide a way to get to the secretstoreclient directly.

By design.  It's an implementation detail.  The secrets bootstrap was intended to encapsulate the idea of getting secrets -- whether via secretstoreclient, from configuration, or from some other method.  

I haven't reviewed this PR and would discourage leaking secretstoreclient (or any other implementation details) outside of the secret bootstrap implementation.

FWIW.  ;)

bnevis-i:
Are third party application vendors expected to use go-mod-bootstrap?  If so, we should be thinking about a way for them to obtain application-specific secrets that doesn't involve forking go-mod-bootstrap to add a custom provider for their particular application.

michaelestrin:
> Are third party application vendors expected to use go-mod-bootstrap?

I can't speak to that.  As far as I know, only go-mod-core-contracts is intended to be an externally consumable module.  

If someone wants to have a custom module, they simply provide a custom bootstrap handler implementation.

lenny-intel:
> Are third party application vendors expected to use go-mod-bootstrap? If so, we should be thinking about a way for them to obtain application-specific secrets that doesn't involve forking go-mod-bootstrap to add a custom provider for their particular application.

@bnevis-i No they are not expected to use it directly. They will use it indirectly via the App and Device SDKs, which each will have the custom provider.

tingyuz:
> * This PR has multiple commits and should be reduced to a single commit.
> * Commit description should follow conventional commits naming style.
> * No unit test for the newly added code
> * Not sure if need to abbreviate Certificate to Cert and PrivateKey to Key...
> 
> Otherwise straightforward and LGTM with the general approach. Did not realize that bootstrap did not provide a way to get to the secretstoreclient directly.

Squash the commits and update the issue commits style. 
The method added is just a wrapper to secretClient.GetSecret(), and the test of latter has been covered in go-mod-secrets, which is the reason there is no unit test case for it provided.

Issue #69, opened 2020-04-17
Make config override env names allow upper case


lenny-intel:
Currently the config override env names as `PascalCase` which is non-standard. Changing the names to UPPERCASE only is a breaking change. For now allow use of UPPERCASE names in addition to the current PascalCase names until release 2.0. At that time we can force only UPPERCASE.

tonyespy:
> Currently the config override env names as PascalCase which is non-standard.

Non-standard by who's definition??? Is this causing breakage somewhere?

I've definitely raised this issue in the past with respect to the fact that there are couple of lowercase environment variables that we currently support. I think I just gave up and didn't mention the camel case when I reviewed the config-seed ADR.

If we're going to convert everything to UPPERCASE in Hanoi, why bother with this now???

lenny-intel:
@tonyespy , no guarantee Hanoi will be 2.0, we all thought Geneva would be 2.0. This allows us to move to uppercase now without it being a backwards breaking change.

PR #70, opened 2020-04-17
feat(environment): Perform case insensitive comparison for config override names

lenny-intel:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Environment variable override names have to be exact case used in TOML file.

Issue Number:  #69

## What is the new behavior?

Environment variable override names can now be uppercase or an case

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/bootstrap/config/environment.go
+++ b/bootstrap/config/environment.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/edgexfoundry/go-mod-configuration/pkg/types"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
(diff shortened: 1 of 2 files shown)


iain-anderson:
OK with this. Because of the way override handling is implemented in device-sdk-c, when I update that it will only accept the original case or all-uppercase but I think that is fine.

lenny-intel:
> OK with this. Because of the way override handling is implemented in device-sdk-c, when I update that it will only accept the original case or all-uppercase but I think that is fine.

I had same thought this weekend that either exact case or upper case only. I have made the tweak to handle this.

bootstrap/config/environment.go
```diff
@@ -116,6 +127,68 @@ func (e *Environment) OverrideConfiguration(lc logger.LoggingClient, serviceConf
...
+			return lcKey, true
+		}
+	}
+
+	return "", false
+}
+
+func isAllUpperCase(key string) bool {
+	for _, ch := range key {
+		if unicode.IsLower(ch) && !unicode.IsUpper(ch) {
```
jim-wang-intel, line 184: Do we need the extra `IsLower(ch)` comparison?  I thought `!Upper` is sufficient...

lenny-intel: That is suppose to be IsLetter. Good catch!!

lenny-intel: Fixed

lenny-intel:
@jpwhitemn , no need to filter out special characters. If they are used, then it just doesn't match... I.e. non of our settings can have special characters. We use underscore because period is not a valid environment character, but the toml path uses period so we have to convert the underscore to a period.

bootstrap/config/environment.go
```diff
@@ -116,6 +127,58 @@ func (e *Environment) OverrideConfiguration(lc logger.LoggingClient, serviceConf
...
+		ucMap[strings.ToUpper(path)] = path
+	}
+
+	return ucMap
+}
+
+// getKeyForMatchedPath searches for match of the environment variable name with the uppercase path (pathMap keys)
+// If matched found to original path (pathMap values) is returned as the "key"
+// For backward compatibility a case insensitive comparision is currently used.
+// TODO: For release v2.0.0 Change this to NOT check that `envVarName` is uppercase and only compare against uppercase
```
rsdmike, line 165: Should we put this as an issue to address in the future?

lenny-intel: Done https://github.com/edgexfoundry/go-mod-bootstrap/issues/71

Issue #85, opened 2020-07-22
Configured startup duration not impacting timeout for service startup


lenny-intel:
Service's pass in a timer that has been initialized with constants for Duration=30s and interval=1s. 

Bootstrap attempts to update the timer with values from '[Startup]' configuration or environment variables, but the update doesn't seem to work.

Recreate by:

1. Start core-data from command-line (no options) without EdgeX stack running.
2. Note that it attempts to connect to the Database approx every second and gives up after approx 30 seconds
3. Edit the local configuration.toml and change  '[Startup]' as follows
```
[Startup]
Duration = 10
Interval = 5
``` 
4. Start core-data from command-line (no options) without EdgeX stack running.
5. Note that it **still** attempts to connect to the Database approx every second and gives up after approx 30 seconds, i.e **no change**

lenny-intel:
The Startup configuration is not useful as it is needed prior to loading the configuration. It will be removed as the values are always 0 when go-mod-bootstrap attempts to use them. The Startup env variables will be the only way to change the startup values from the default values.

PR #86, opened 2020-07-24
fix: Startup Duration and Interval never updated from default values

lenny-intel:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Changing the Startup configuration has no impact and neither does using the startup env variables.  
The Startup configuration is not useful as it is needed prior to loading the configuration. 
Update of timer with values from Env Vars doesn't work.

Issue Number: #85

## What is the new behavior?

Service Startup config is not loaded when timer is created/updated so have been removed.
NewStartUpTimer now checks and uses Env values form start so update of timer no longer needed.

- [ ] Yes
- [x] No

Startup configuration never worked, so removing it will is not a breaking change

## Are there any new imports or modules? If so, what are they used for and why?
No

## Are there any specific instructions or things that should be known prior to reviewing?

Code restructuring was required to avoid circular imports
Requires changes to each service for the change in signature of NewStartUpTimer 

## Other information

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright 2019 Dell Inc.
+ * Copyright 2020 Intel Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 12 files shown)


lenny-intel:
> I did not test all envVar overrides

@jpwhitemn , if you desire..  you can the start-up env overrides by running service like core-data (built with my branch of go-mod-bootstrap) w/o edgex stack running using this command:
```
EDGEX_STARTUP_DURATION=120 EDGEX_STARTUP_INTERVAL=30 ./core-data
```
Verify it took ~120 secs (rather than ~30 secs) to abort and only retried every ~30 secs (rather than ~1 secs)

PR #125, opened 2020-11-18
feat: Allow service to pass in initial logging client

lenny-intel:
## PR Checklist

Initial Logging client created by boot-strap uses DEBUG level and can not be overridden. This client is used until configuration is loaded.

## What is the new behavior?
Service can now pass an initial Logging Client in the DIC for bootstrap to use instead of creating one.
This enables Security Utility to run in silent mode as initial client can be set to EROOR instead of DEBUG as it now.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -17,6 +17,7 @@ package bootstrap
 
 import (
 	"context"
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/logging"
 	"os"
 	"os/signal"
 	"sync"
@@ -30,7 +31,6 @@ import (
 	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/environment"
 	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/flags"
 	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/interfaces"
-	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/logging"
 	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/registration"
 	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/startup"
 	"github.com/edgexfoundry/go-mod-bootstrap/di"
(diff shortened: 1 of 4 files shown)


bootstrap/bootstrap.go
```diff
@@ -91,7 +91,11 @@ func RunAndReturnWaitGroup(
 	var wg sync.WaitGroup
 	deferred := func() {}
 
-	lc := logging.FactoryToStdout(serviceKey)
+	// Check is service provided an initial Logging Client to use. If not create one.
+	lc := container.LoggingClientFrom(dic.Get)
+	if lc == nil {
```
jim-wang-intel, line 96: maybe absorb this logic into `container.LoggingClientFrom(dic.Get)` above?

go.mod
```diff
@@ -11,4 +11,7 @@ require (
 	github.com/stretchr/testify v1.5.1
 )
 
+replace (
+	github.com/edgexfoundry/go-mod-core-contracts => ../go-mod-core-contracts
+)
```
bnevis-i, line 16: This shouldn't have been in the commit?

lenny-intel: Was needed for testing. Left in as a reminder that core-contracts version will need to be bumped when that PR is merged.

bootstrap/bootstrap.go
lenny-intel: Nope, the get is simple retrieve from DIC. Should not have a side effect.

jim-wang-intel: Not sure what you meant. why the reason is something to do with dic.Get? the check is purely on `LoggingClientFrom` function.

lenny-intel: That is how we retrieve from the DIC. Syntax is a bit strange... ;-)

bnevis-i (approved):

bnevis-i:
@lenny-intel Test cases broken.

bnevis-i (approved):

jpwhitemn (approved):

PR #216, opened 2021-04-20
fix: Replace hyphen with underscore in override names

lenny-intel:
## PR Checklist

Hyphens accepted on override names
Hyphen is not a standard character for environment variable names

## What is the new behavior?
Hyphens are replaced with underscore in override names

- [x] Yes
- [ ] No

BREAKING CHANGE: Overrides that have hyphens will not longer work and must be updated to replace hyphens with underscores.

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/bootstrap/environment/variables.go
+++ b/bootstrap/environment/variables.go
@@ -61,12 +61,7 @@ type Variables struct {
 	lc        logger.LoggingClient
 }
 
-type overrideSpec struct {
-	tomlPath     string
-	overrideName string
-}
-
-// NewEnvironment constructor reads/stores os.Environ() for use by Variables receiver methods.
+// NewVariables constructor reads/stores os.Environ() for use by Variables receiver methods.
 func NewVariables(lc logger.LoggingClient) *Variables {
 	osEnv := os.Environ()
 	e := &Variables{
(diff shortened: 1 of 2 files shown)


jim-wang-intel (approved):
LGTM

bootstrap/environment/variables.go
```diff
@@ -163,24 +170,31 @@ func (e *Variables) buildPaths(keyMap map[string]interface{}) []string {
 	return paths
 }
 
-// buildUppercasePathMap builds a map where the key is the uppercase version of the path
-// and the value is original version of the path
-func (e *Variables) buildUppercasePathMap(paths []string) map[string]string {
-	ucMap := make(map[string]string)
+func (e *Variables) buildOverrideSpecs(paths []string) []overrideSpec {
```
cloudxxx8, line 173: why not keep using `map[string]string`?  It could remove `getPathForMatchedOverride` and make sure the overrideName (key) is unique.

In line 126, it could be

`path, found := overrideSpecMap[envVar]`

lenny-intel: Very good point. THX!

lenny-intel: Done. See https://github.com/edgexfoundry/go-mod-bootstrap/pull/216/commits/710eb7ce820196f535f72194ecdb7dcd18992be6

iain-anderson (approved):
LGTM

PR #246, opened 2021-05-25
fix: Add conditional for error message and return false on error

lenny-intel:
## PR Checklist

Error message is log when there isn't an error. When the is an error the handler still returns true.

## What is the new behavior?
Error message is only logged when there is an error and hander returns false on error.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


cloudxxx8 (approved):
LGTM

Issue #249, opened 2021-05-26
[Tech Debt] Add EdgeX internal secret path prefix for SecretStore config Path property


jim-wang-intel:
To reduce the duplicates of `/v1/secret/edgex/`in `Path` property of SecretStore config element, that path can be refactored into internal code and prefix it with the **service specific** secret path like `core-command`.

Note that this is a breaking change and all services with `SecretStore` config in TOML will need to be changed.

PR #250, opened 2021-05-26
refactor!: Add internal prefix for secret path on SecretStore config

jim-wang-intel:
Add internal prefix `/v1/secret/edgex/` for secret's Path in `SecretStore` config so that now we only need to specify the service specific secret path instead of the full path with /v1/secret/edgex/ + serviceSecrets

Closes: #249

Signed-off-by: Jim Wang <yutsung.jim.wang@intel.com>

## PR Checklist

- [x ] Tests for the changes have been added (for bug fixes / features)

Currently the `Path` specified in the `SecretStore` config is full path with /v1/secret/edgex/ prefixed in all services using secret store.

## What is the new behavior?
The fixed prefix `/v1/secret/edgex/` now is added and prefixed in the code implementation and now for every service just needs to specify the service specific secret path.

- [x ] Yes
- [ ] No

**Note: this change will impact all services using SecretStore config and thus require changes in toml files.**

## New Imports

- [ ] Yes
- [x ] No

## Specific Instructions

## Other information

### DIFF
--- a/bootstrap/secret/secret.go
+++ b/bootstrap/secret/secret.go
@@ -1,5 +1,6 @@
 /********************************************************************************
  *  Copyright 2019 Dell Inc.
+ *  Copyright 2021 Intel Corp.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


lenny-intel (approved):
LGTM

PR #276, opened 2021-09-22
feat: Add optional capability to seed service secrets

### LINKED ISSUE #273: Add capability to seed unknown secrets during bootstrap
### Description
Currently `unknown` secrets, those know only at deployment time for things like camera credentials, external broker credentials,  HTTP API keys, etc, have to be pushed into the service Secret Store via the `/secret` [REST endpoint](https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/app-functions-sdk/2.0.0#/default/post_secret) once the service is up and running. This can be problematic if the service needs the credentials immediately after bootstraping, i.e. Device MQTT connecting to broker or Device Camera connecting the camera(s) as the devices are added. For add-on services the known secrets are specified to be added via the `ADD_KNOWN_SECRETS` environment variable ([See here for more details](https://docs.edgexfoundry.org/2.0/security/Ch-Configuring-Add-On-Services/#optional-configure-known-secrets-for-add-on-services)).

See related closed issue for more background discussion on this new capability: https://github.com/edgexfoundry/edgex-go/issues/3709

We need an alternative to the REST endpoint were the deployment can specify the unknown secrets to be populated during bootstrapping so the the secrets are available when the service needs them.

This can be accomplished with a file supplied to the service that contains the secrets to be added to the service's `SecretStore`. The file will be root access only. For docker, it will be volume mounted into the service's container. For snaps, the file can be provided via a combination of configure hook and/or content interface.

An new `SecretsFile` setting will be added to the `[SecretStore]` configuration in use by all services. This is the local path to the secrets file. By default this setting is empty. If set, the file will be read, after the service SecretProvider is created, and the secrets will be push into the service's SecretStore. 

This file will contain JSON that is an array off objects that contain the secret path, data and flag indicating if  the secret data has been imported. The data is similar to that  POSTed to the `/secrets` endpoint, but without the extra request DTO information and the addition of `imported`.

Example:
```
{
    "secrets": [
        {
            "path": "credentails001",
            "imported": false,
            "secretData": [
                {
                    "key": "user1",
                    "value": "password1"
                }
            ]
        },
        {
            "path": "credentails002",
            "imported": false,
            "secretData": [
                {
                    "key": "user2",
                    "value": "password2"
                }
            ]
        }
    ]

```

Once the secrets have been "imported" the file will be written over with the `imported` flags set to true and the secret data removed.

Example file after imported:

```
{
    "secrets": [
        {
            "path": "credentails001",
            "imported": true,
            "secretData": []
        },
        {
            "path": "credentails002",
            "imported": true,
            "secretData": []
        }
    ]
```

lenny-intel:
Secrets are seeded from a JSON file specified by the SecretStore.SecretsFile setting
If SecretsFile setting is blank, seeding is skipped.

closes #273

This PR depends on https://github.com/edgexfoundry/go-mod-secrets/pull/126 to be merged first.

## PR Checklist

  https://github.com/edgexfoundry/edgex-docs/pull/587

## Testing Instructions

1. Start edgex using compose builder `make run ds-camera`
   - This will create the `SecretStore` for device camera
2. Stop the Device Camera container
1. Clone branch for this PR and branch for go-mod-secrets PR  https://github.com/edgexfoundry/go-mod-secrets/pull/126
2. Add the following to device-sdk-go `go.mod`

```
replace (
	github.com/edgexfoundry/go-mod-bootstrap/v2 => ../go-mod-bootstrap
)
```
3. Add following the device-camera-go `go.mod`
```
replace (
	github.com/edgexfoundry/device-sdk-go/v2 => ../device-sdk-go
	github.com/edgexfoundry/go-mod-bootstrap/v2 => ../go-mod-bootstrap
)
```
4. Build device-camera
5. cd to `cmd` folder in device-camera-go
6. create file `/tmp/camera-secrets.json` with the following JSON:
    ```json
    {
        "secrets": [
            {
                "path": "credentials001",
                "imported": false,
                "secretData": [
                    {
                        "key": "username",
                        "value": "my-user-1"
                    },
                                    {
                        "key": "password",
                        "value": "password-001"
                    }
                ]
            }
        ]
    }
    ```
7. run `sudo EDGEX_SECURITY_SECRET_STORE=true SECRETSTORE_SECRETSFILE=/tmp/camera-secrets.json ./device-camera`
8. Device Camera logs will have the following:
```
level=INFO ts=2021-09-27T18:22:23.1861784Z app=device-camera source=secret.go:52 msg="Creating SecretClient"
level=INFO ts=2021-09-27T18:22:23.1862282Z app=device-camera source=secret.go:59 msg="Reading secret store configuration and authentication token"
level=INFO ts=2021-09-27T18:22:23.1878451Z app=device-camera source=secret.go:71 msg="Attempting to create secret client"
level=INFO ts=2021-09-27T18:22:23.2092923Z app=device-camera source=secrets.go:277 msg="kick off token renewal with interval: 30m0s"
level=INFO ts=2021-09-27T18:22:23.209413Z app=device-camera source=secure.go:229 msg="Seeding 1 Service Secrets"
level=INFO ts=2021-09-27T18:22:23.2115168Z app=device-camera source=secure.go:248 msg="Secret for 'credentials001' successfully stored."
level=INFO ts=2021-09-27T18:22:23.2117384Z app=device-camera source=secure.go:218 msg="Scrubbing of secrets file complete."
level=INFO ts=2021-09-27T18:22:23.2117899Z app=device-camera source=secret.go:88 msg="Created SecretClient"
``` 
**And will `NOT` display any error getting secrets, but will fail to initialize the camera since no camera exists.**
9. The `/tmp/camera-secrets.json` file contents are now:
```
"secrets":[{"path":"credentials001","imported":true,"secretData":[]}]}
```
10. recreate `/tmp/camera-secrets.json` file contents as above with secret data.
11. run `sudo EDGEX_SECURITY_SECRET_STORE=true SECRETSTORE_SECRETSFILE=/tmp/camera-secrets.json SECRETSTORE_DISABLESCRUBSECRETSFILE=true ./device-camera`
12. Device Camera logs will have the following additional message:
```
level=INFO ts=2021-09-27T18:18:12.3870583Z app=device-camera source=secure.go:210 msg="Scrubbing of secrets file disable."
```
13. The `/tmp/camera-secrets.json` file contents will not be changed. I.e. secret data will still be present in the file

## New Dependency Instructions (If applicable)
**N/A**

### DIFF
--- a/bootstrap/secret/secret.go
+++ b/bootstrap/secret/secret.go
@@ -76,7 +76,7 @@ func NewSecretProvider(
 					lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
 
 					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
-						err = secureProvider.LoadServiceSecrets(secretStoreConfig.SecretsFile, secretStoreConfig.DisableScrubSecretsFile)
+						err = secureProvider.LoadServiceSecrets(secretStoreConfig)
 						if err != nil {
 							return nil, err
 						}
--- a/bootstrap/secret/secret.go
+++ b/bootstrap/secret/secret.go
@@ -72,6 +72,18 @@ func NewSecretProvider(
 				secretClient, err = secrets.NewSecretsClient(ctx, secretConfig, lc, secureProvider.DefaultTokenExpiredCallback)
 				if err == nil {
 					secureProvider.SetClient(secretClient)
+
+					lc.Infof("SecretsFile is '%s'", secretConfig.SecretsFile)
+
+					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
+						err = secureProvider.LoadServiceSecrets(secretConfig.SecretsFile)
+						if err != nil {
+							return nil, err
+						}
+					} else {
+						lc.Infof("SecretsFile not set, skipping seeding of service secrets.")
+					}
+
 					provider = secureProvider
 					lc.Info("Created SecretClient")
 					break
@@ -107,6 +119,7 @@ func getSecretConfig(secretStoreInfo config.SecretStoreInfo, tokenLoader authtok
 		Host:           secretStoreInfo.Host,
 		Port:           secretStoreInfo.Port,
 		Path:           addEdgeXSecretPathPrefix(secretStoreInfo.Path),
+		SecretsFile:    secretStoreInfo.SecretsFile,
 		Protocol:       secretStoreInfo.Protocol,
 		Namespace:      secretStoreInfo.Namespace,
 		RootCaCertPath: secretStoreInfo.RootCaCertPath,
--- a/bootstrap/secret/secret.go
+++ b/bootstrap/secret/secret.go
@@ -72,20 +72,20 @@ func NewSecretProvider(
 				secretClient, err = secrets.NewSecretsClient(ctx, secretConfig, lc, secureProvider.DefaultTokenExpiredCallback)
 				if err == nil {
 					secureProvider.SetClient(secretClient)
+					provider = secureProvider
+					lc.Info("Created SecretClient")
 
 					lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
 
-					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
-						err = secureProvider.LoadServiceSecrets(secretStoreConfig)
-						if err != nil {
-							return nil, err
-						}
-					} else {
+					if len(strings.TrimSpace(secretConfig.SecretsFile)) == 0 {
 						lc.Infof("SecretsFile not set, skipping seeding of service secrets.")
+						break
 					}
 
-					provider = secureProvider
-					lc.Info("Created SecretClient")
+					err = secureProvider.LoadServiceSecrets(secretStoreConfig)
+					if err != nil {
+						return nil, err
+					}
 					break
 				}
 			}
--- a/bootstrap/secret/secret.go
+++ b/bootstrap/secret/secret.go
@@ -76,7 +76,7 @@ func NewSecretProvider(
 					lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
 
 					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
-						err = secureProvider.LoadServiceSecrets(secretConfig.SecretsFile)
+						err = secureProvider.LoadServiceSecrets(secretStoreConfig.SecretsFile, secretStoreConfig.DisableScrubSecretsFile)
 						if err != nil {
 							return nil, err
 						}
--- a/bootstrap/secret/secret.go
+++ b/bootstrap/secret/secret.go
@@ -73,7 +73,7 @@ func NewSecretProvider(
 				if err == nil {
 					secureProvider.SetClient(secretClient)
 
-					lc.Infof("SecretsFile is '%s'", secretConfig.SecretsFile)
+					lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
 
 					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
 						err = secureProvider.LoadServiceSecrets(secretConfig.SecretsFile)
--- a/bootstrap/secret/secure.go
+++ b/bootstrap/secret/secure.go
@@ -197,6 +197,7 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
 	return reReadToken, true
 }
 
+// LoadServiceSecrets loads the service secrets from the specified file and stores them in the service's SecretStore
 func (p *SecureProvider) LoadServiceSecrets(filePath string) error {
 	contents, err := os.ReadFile(filePath)
 	if err != nil {
--- a/bootstrap/secret/secure.go
+++ b/bootstrap/secret/secure.go
@@ -198,7 +198,7 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
 }
 
 // LoadServiceSecrets loads the service secrets from the specified file and stores them in the service's SecretStore
-func (p *SecureProvider) LoadServiceSecrets(filePath string) error {
+func (p *SecureProvider) LoadServiceSecrets(filePath string, scrubDisabled bool) error {
 	contents, err := os.ReadFile(filePath)
 	if err != nil {
 		return fmt.Errorf("seeding secrets failed: %s", err.Error())
@@ -206,9 +206,16 @@ func (p *SecureProvider) LoadServiceSecrets(filePath string) error {
 
 	data, seedingErrs := p.seedSecrets(contents)
 
+	if scrubDisabled {
+		p.lc.Infof("Scrubbing of secrets file disable.")
+		return seedingErrs
+	}
+
 	if err := os.WriteFile(filePath, data, 0); err != nil {
 		return fmt.Errorf("seeding secrets failed: unable to overwrite file with secret data removed: %s", err.Error())
 	}
+
+	p.lc.Infof("Scrubbing of secrets file complete.")
 
 	return seedingErrs
 }
--- a/bootstrap/secret/secure.go
+++ b/bootstrap/secret/secure.go
@@ -27,6 +27,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/interfaces"
+	"github.com/edgexfoundry/go-mod-bootstrap/v2/config"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/v2/clients/logger"
 
@@ -198,20 +199,21 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
 }
 
 // LoadServiceSecrets loads the service secrets from the specified file and stores them in the service's SecretStore
-func (p *SecureProvider) LoadServiceSecrets(filePath string, scrubDisabled bool) error {
-	contents, err := os.ReadFile(filePath)
+func (p *SecureProvider) LoadServiceSecrets(secretStoreConfig config.SecretStoreInfo) error {
+
+	contents, err := os.ReadFile(secretStoreConfig.SecretsFile)
 	if err != nil {
 		return fmt.Errorf("seeding secrets failed: %s", err.Error())
 	}
 
 	data, seedingErrs := p.seedSecrets(contents)
 
-	if scrubDisabled {
+	if secretStoreConfig.DisableScrubSecretsFile {
 		p.lc.Infof("Scrubbing of secrets file disable.")
 		return seedingErrs
 	}
 
-	if err := os.WriteFile(filePath, data, 0); err != nil {
+	if err := os.WriteFile(secretStoreConfig.SecretsFile, data, 0); err != nil {
 		return fmt.Errorf("seeding secrets failed: unable to overwrite file with secret data removed: %s", err.Error())
 	}
 
--- a/bootstrap/secret/secure.go
+++ b/bootstrap/secret/secure.go
@@ -18,12 +18,18 @@ package secret
 import (
 	"errors"
 	"fmt"
+	"os"
+	"strings"
 	"sync"
 	"time"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/dtos/common"
+	"github.com/hashicorp/go-multierror"
 
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/interfaces"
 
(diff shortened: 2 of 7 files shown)


siggiskulason:
LGTM. Thanks for the testing instructions. I tested this with my ONVIF camera, the edgexfoundry snap and a natively running device-camera, setting the following in the camera `configuration.toml` file:

```
SecretsFile = "/home/sidar/tmp/camera-secrets.json"
TokenFile = "/var/snap/edgexfoundry/current/secrets/device-camera/secrets-token.json"
```
with `camera-secrets.json` being

```
{
    "secrets": [
        {
            "path": "credentials001",
            "imported": false,
            "secretData": [
                {
                    "key": "username",
                    "value": "admin"
                },
                                {
                    "key": "password",
                    "value": "admin"
                }
            ]
        }
    ]
}

```

and then running with

```
sudo EDGEX_SECURITY_SECRET_STORE=true ./device-camera
```

that works fine:
- The secrets are seeded correctly
- the JSON file gets overwritten as expected
- The service authenticates with the camera and I see readings appear with images.

siggiskulason (approved):

lenny-intel:
> I tested this with my ONVIF camera, the edgexfoundry snap and a natively running device-camera,

@siggiskulason Excellent!! Thanks!

bootstrap/secret/secret.go
```diff
@@ -72,6 +72,18 @@ func NewSecretProvider(
 				secretClient, err = secrets.NewSecretsClient(ctx, secretConfig, lc, secureProvider.DefaultTokenExpiredCallback)
 				if err == nil {
 					secureProvider.SetClient(secretClient)
+
+					lc.Infof("SecretsFile is '%s'", secretConfig.SecretsFile)
```
jim-wang-intel, line 76: would suggest putting this in debug level instead of info as it is kind of sensitive

lenny-intel: Updated

jim-wang-intel (approved):
LGTM

bootstrap/secret/secure.go
```diff
@@ -190,3 +196,69 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
 
 	return reReadToken, true
 }
+
+func (p *SecureProvider) LoadServiceSecrets(filePath string) error {
```
tonyespy, line 200: Should there be godoc for this method?

bootstrap/secret/secure.go
```diff
@@ -190,3 +196,69 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
...
+	contents, err := os.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("seeding secrets failed: %s", err.Error())
+	}
+
+	data, seedingErrs := p.seedSecrets(contents)
+
+	if err := os.WriteFile(filePath, data, 0); err != nil {
+		return fmt.Errorf("seeding secrets failed: unable to overwrite file with secret data removed: %s", err.Error())
```
tonyespy, line 209: Has seeding secrets really failed in this case? It seems to me its possible for p.seedSecrets to succeed and actually provision the supplied secrets, yet if the file isn't writable for whatever reason, we tell the calling service that the operation failed.

Maybe you should just log a warning in the case of the write failure?

bootstrap/secret/secure.go
```diff
@@ -190,3 +196,69 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
...
+func prepareSecret(secret ServiceSecret) (string, map[string]string) {
+	var secretsKV = make(map[string]string)
+	for _, secret := range secret.SecretData {
+		secretsKV[secret.Key] = secret.Value
+	}
+
+	path := strings.TrimSpace(secret.Path)
+
+	return path, secretsKV
+}
```
tonyespy, line 264: missing NL @ EOF

tonyespy (commented):
A couple of inline comments...

bootstrap/secret/secure.go
lenny-intel: Yes

bootstrap/secret/secure.go
lenny-intel: HUmm, i get where you are coming from, but I think @bnevis-i would prefer this to be an error since it didn't rewrite the file w/o the secretData. No one will see the warning message since the secrets get seeded.

bootstrap/secret/secure.go
lenny-intel: Updated

bootstrap/secret/secure.go
lenny-intel: added

bootstrap/secret/secure.go
tonyespy: Should we also remove the secrets then too (if possible) so that we're handling this as a single transaction?

lenny-intel: > Should we also remove the secrets then too (if possible) so that we're handling this as a single transaction?

No, it doesn't hurt to store them again once the write issue is resolved.

bnevis-i (commented):
Please tell us about the new dependency.

bootstrap/secret/secure.go
lenny-intel: @bnevis-i , please comment on error and fail vs. logging warning when writing file back out w/o the secret data.

lenny-intel:
> Please tell us about the new dependency.

@bnevis-i , it is new to this module, but not EdgeX. It is on the approved list here:
https://wiki.edgexfoundry.org/display/FA/Vetting+Process+for+3rd+Party+Dependencies

bootstrap/secret/secure.go
bnevis-i: Let's:
 - If secrets are there, import them
 - Attempt to scrub them unless told not to
 - Fail if either can't import or can't scrub when we're told to.

lenny-intel: @tonyespy , @bnevis-i , I am adding the `DisableScrubSecretsFile` setting to SecretStoreInfo to allow users to control this. Default will be false so secrets get scrubbed unless they explicitly change the setting.

bootstrap/secret/secret.go
```diff
@@ -76,7 +76,7 @@ func NewSecretProvider(
 					lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
 
 					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
-						err = secureProvider.LoadServiceSecrets(secretConfig.SecretsFile)
+						err = secureProvider.LoadServiceSecrets(secretStoreConfig.SecretsFile, secretStoreConfig.DisableScrubSecretsFile)
```
jim-wang-intel, line 79: Now we have two params from the same instance of  `secretStoreConfig`.  would like to suggest refactoring this to just pass in `secretStoreConfig` instance so that the logic of dependency is hidden behind that func and also easier to extend later on if there are more.

lenny-intel: Sure

lenny-intel: Done

bootstrap/secret/secure.go
```diff
@@ -190,3 +197,78 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
...
+	if err := os.WriteFile(secretStoreConfig.SecretsFile, data, 0); err != nil {
+		return fmt.Errorf("seeding secrets failed: unable to overwrite file with secret data removed: %s", err.Error())
+	}
+
+	p.lc.Infof("Scrubbing of secrets file complete.")
+
+	return seedingErrs
+}
+
+func (p *SecureProvider) seedSecrets(contents []byte) ([]byte, error) {
```
bnevis-i, line 225: Would it be cleaner to take a data structure here rather than a byte array?  Or a pointer to the data structure that you modify directly?

bootstrap/secret/secure.go
```diff
@@ -190,3 +197,78 @@ func (p *SecureProvider) DefaultTokenExpiredCallback(expiredToken string) (repla
 
 	return reReadToken, true
 }
+
+// LoadServiceSecrets loads the service secrets from the specified file and stores them in the service's SecretStore
+func (p *SecureProvider) LoadServiceSecrets(secretStoreConfig config.SecretStoreInfo) error {
+
+	contents, err := os.ReadFile(secretStoreConfig.SecretsFile)
+	if err != nil {
+		return fmt.Errorf("seeding secrets failed: %s", err.Error())
```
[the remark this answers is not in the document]
lenny-intel: Because it the resulting log message is unreadable.

bootstrap/secret/secure.go
lenny-intel: This is the contents just read from the file, which this func marshals into struct

bnevis-i (approved):

bootstrap/secret/secret.go
```diff
@@ -72,6 +72,18 @@ func NewSecretProvider(
 				secretClient, err = secrets.NewSecretsClient(ctx, secretConfig, lc, secureProvider.DefaultTokenExpiredCallback)
 				if err == nil {
 					secureProvider.SetClient(secretClient)
+
+					lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
+
+					if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
```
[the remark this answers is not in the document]
lenny-intel: How? still an if/else just reversing the logic.

jim-wang-intel (approved):
LGTM

bootstrap/secret/secret.go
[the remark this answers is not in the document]
bnevis-i:
```
  secureProvider.SetClient(secretClient)
  lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
  if len(strings.TrimSpace(secretConfig.SecretsFile)) > 0 {
    err = secureProvider.LoadServiceSecrets(secretStoreConfig)
    if err != nil {
      return nil, err
    }
  } else {
    lc.Infof("SecretsFile not set, skipping seeding of service secrets.")
  }
  provider = secureProvider


  secureProvider.SetClient(secretClient)
  lc.Debugf("SecretsFile is '%s'", secretConfig.SecretsFile)
  if len(strings.TrimSpace(secretConfig.SecretsFile)) == 0 {
    lc.Infof("SecretsFile not set, skipping seeding of service secrets.")
    break
  err = secureProvider.LoadServiceSecrets(secretStoreConfig)
  if err != nil {
    return nil, err
  }
  provider = secureProvider
```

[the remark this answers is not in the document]
lenny-intel: Yep! :-)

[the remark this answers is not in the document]
lenny-intel: I am always an advocate of early return to avoid indentation, just didn't see it here. THX!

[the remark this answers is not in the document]
lenny-intel: Opps, not so fast. This skips the following important code when you do the `break` that you added, so still need the if/else of have repeated code.

```
					provider = secureProvider
					lc.Info("Created SecretClient")
```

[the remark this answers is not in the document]
lenny-intel: OK, was able to move that code up before handling of the secrets file and now the break you added will work.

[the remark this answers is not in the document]
lenny-intel: Updated

bnevis-i (approved):

jim-wang-intel (approved):

tonyespy (commented):
I just want to clarify what's expected if the scrub fails, should the secrets provisioned into vault also be scrubbed? You also could structure the code such that the write to vault doesn't happen until after the JSON file gets scrubbed...  If @bnevis-i you're OK with the way this is implemented, then I'll just approve, but as implemented this isn't being handled as a single transaction.

bnevis-i:
> I just want to clarify what's expected if the scrub fails, should the secrets provisioned into vault also be scrubbed? You also could structure the code such that the write to vault doesn't happen until after the JSON file gets scrubbed... If @bnevis-i you're OK with the way this is implemented, then I'll just approve, but as implemented this isn't being handled as a single transaction.

Yes, if the scrub fails the secrets will still be in the secret store.  It is impossible to truly have a single transaction here as neither Vault nor the file system support a two-phase commit protocol.  I'd rather have this behavior than the possibility that the JSON is scrubbed and then Vault fails, because there is no recovery from that.

lenny-intel:
> I just want to clarify what's expected if the scrub fails, should the secrets provisioned into vault also be scrubbed? You also could structure the code such that the write to vault doesn't happen until after the JSON file gets scrubbed... If @bnevis-i you're OK with the way this is implemented, then I'll just approve, but as implemented this isn't being handled as a single transaction.

@tonyespy, The secrets do not need to be scrubbed from vault if file write fails. No harm in them still being there. Once write issue is resolved they will be over written. Trying to scrub them out of vault will add unneeded complexity.

PR #327, opened 2022-04-25
feat: Enhance metric name matching to match on prefix

lenny-intel:
This allows for a class of Metrics to be enabled with one configured
metric name. App SDK uses this for PipelineMetrics by appending the
pipeline ID to the name of the metric(s) it is collecting for multiple
function pipelines.

Signed-off-by: Leonard Goodell <leonard.goodell@intel.com>

## PR Checklist

## Testing Instructions
Run non-secure EdgeX stack and stop Core Data container
Use this branch to build Core Data locally
Change Metric config to enable both metrics
Run with DEBUG logging and verify 2 metrics are reported.

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -254,6 +254,6 @@ func (t *TelemetryInfo) MetricEnabled(metricName string) bool {
 		return enabled
 	}
 
-	// Service's metric  name did match any config Metric name.
+	// Service's metric name did not match any config Metric name.
 	return false
 }
--- a/config/types.go
+++ b/config/types.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


config/types.go
```diff
@@ -240,11 +241,19 @@ type TelemetryInfo struct {
...
+		// of the metric(s) it is collecting for multiple function pipelines.
+		if !strings.HasPrefix(metricName, configMetricName) {
+			continue
+		}
+
+		return enabled
 	}
 
-	return enabled
+	// Service's metric  name did match any config Metric name.
```
jim-wang-intel, line 257: did `not` match?

jim-wang-intel (approved):

PR #362, opened 2022-07-22
fix: Run WatchForChange in a new thread

### LINKED ISSUE #361: Error from WatchForChanges is not received
# 🐞 Bug Report

### Description 
The error returned from `WatchForChanges` is asynchronous and sent by the `errorStream` channel.

Any error sent in the main thread of line 461 cannot be received in the case of line 470 as the `for` loop is executed in the same thread after `WatchForChanges`.
https://github.com/edgexfoundry/go-mod-bootstrap/blob/aaf2123c3c853951119ee9789949c8c291d3792d/bootstrap/config/config.go#L461-L471

`configClient.WatchForChanges(updateStream, errorStream, serviceConfig.EmptyWritablePtr(), writableKey) ` needs to be executed in another thread.

lindseysimple:
fixes #361.
Run WatchForChange in a new thread to receive the error chan.

Signed-off-by: Lindsey Cheng <beckysocute@gmail.com>

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) N/A
- [ ] I have opened a PR for the related docs change (if not, why?) N/A, it's the internal behavior
  <link to docs PR>

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -458,7 +458,7 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
 		updateStream := make(chan interface{})
 		defer close(updateStream)
 
-		configClient.WatchForChanges(updateStream, errorStream, serviceConfig.EmptyWritablePtr(), writableKey)
+		go configClient.WatchForChanges(updateStream, errorStream, serviceConfig.EmptyWritablePtr(), writableKey)
 
 		for {
 			select {
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -468,7 +468,7 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
 				return
 
 			case ex := <-errorStream:
-				lc.Error(ex.Error())
+				lc.Errorf("error occurred during listening to the configuration changes: %s", ex.Error())
 
 			case raw, ok := <-updateStream:
 				if !ok {


bootstrap/config/config.go
```diff
@@ -458,7 +458,7 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
 		updateStream := make(chan interface{})
 		defer close(updateStream)
 
-		configClient.WatchForChanges(updateStream, errorStream, serviceConfig.EmptyWritablePtr(), writableKey)
+		go configClient.WatchForChanges(updateStream, errorStream, serviceConfig.EmptyWritablePtr(), writableKey)
```
lenny-intel, line 461: This is redundant.
The implementation of configClient.WatchForChanges is already async. 
https://github.com/edgexfoundry/go-mod-configuration/blob/main/internal/pkg/consul/client.go#L250

cloudxxx8: @lenny-intel this is calling the abstraction interface fucnction, so we don't know whether the implementation is async.
also if something wrong happens in this area, the process will be stuck
https://github.com/edgexfoundry/go-mod-configuration/blob/a360551635e96eb79b5c7c3dd45f86fa62f7369c/internal/pkg/consul/client.go#L235-L245

From my opinion, I thought the redundant is in the consul/client.go 

In our use case, we would like to establish the message bus connection first, and there is no way to return the error if the connection cannot be established.

lenny-intel (approved):
LGTM

Issue #349, opened 2022-06-16
Ability to register a callback for when SecretStore has been modified


ajcasagrande:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

go-mod-bootstrap SecretProvider and SecretStore

### Description [**REQUIRED**]
Similar to how you can register a callback to be notified when a service's configuration is modified from the configuration provider:
https://github.com/edgexfoundry/go-mod-bootstrap/blob/010839c6db32356d141c0ae689095f3ceef96608/bootstrap/config/config.go#L290-L293

it would be nice to have the same functionality when the SecretStore is updated. It would need to be able to be registered from an app-service or device-service similar to the config ability.

The advantages of this is that it can also cut down or remove the need for writing code that utilizes logic like this: #343

Using the `device-onvif-camera` as an example, when a secret is added or updated, the callback can check which Cameras are associated with that secret and modify the client object to use that new value in realtime. This allows you to cache the client object without fear that the credentials are outdated. To replicate the same functionality you would need to lookup the secret value every time you want to use the client.

### Describe the solution you'd like
Ability to register a callback with the SecretProvider that gets called anytime a secret is added, updated, or removed. Ideally it would provide you with some details, such as the name/path of what changed, but this is optional.

### Describe alternatives you've considered
Custom polling solution.

lenny-intel:
Since this is on the running service, we should be able to do this in the Secret Provider. i.e when SetSecret() is called successfully, call any callbacks that are registered.

bnevis-i:
I don't know Anthony's take, but personally, I don't think doing this all in-process would be very useful.  It seems that this functionality would be needed when an external process modifies the secret, and you want your process to be notified.

Sadly, Vault does not have such functionality and we could only get it by polling a specific secret.  For the same reason that secret enumeration isn't available, we also wouldn't be able to get an "anything changed" notification either.

As an alternative implementation, I suggest the following:

- store an encryption key in the secret store
- write your connection configurations to a directly on the file system, encrypting it with the above key first
- watch the file system for changes using fsnotify
- load new configuration for the affected files

Such a change would also allow full enumeration of the secrets as well.

This solution is somewhat kubernetes unfriendly, as it would require availability of RWX (shared filesystem) volume types in contrast to the easier RWO (exclusive use) volume types.

bnevis-i:
[...]

PR #373, opened 2022-09-20
feat: Added logic for secret update callback

drkfmorton:
Fixes https://github.com/edgexfoundry/go-mod-bootstrap/issues/349

## PR Checklist

  <link to docs PR>


drkfmorton:
Created this draft to get feedback as I complete testing and unit testing.

bootstrap/interfaces/secret.go
```diff
@@ -28,4 +28,10 @@ type SecretProvider interface {
 
 	// HasSecret returns true if the service's SecretStore contains a secret at the specified path.
 	HasSecret(path string) (bool, error)
+
+	// RegisteredSecretUpdateCallback registers a callback for a secret.
+	RegisteredSecretUpdateCallback(path string, callback func(path string))
+
+	// SecretsUpdatedWithPath .
+	SecretsUpdatedWithPath(path string)
```
bnevis-i, line 36:
```suggestion
	SecretUpdatedAtPath(path string)
```

However, should this really be a public interface method?  Isn't this internal detail?  Why would a user of the interface need to call this method?

bootstrap/interfaces/secret.go
```diff
@@ -28,4 +28,10 @@ type SecretProvider interface {
 
 	// HasSecret returns true if the service's SecretStore contains a secret at the specified path.
 	HasSecret(path string) (bool, error)
+
+	// RegisteredSecretUpdateCallback registers a callback for a secret.
+	RegisteredSecretUpdateCallback(path string, callback func(path string))
```
bnevis-i, line 33: Would it be useful for the callback to actually supply the callee with the new secret value?

```suggestion
	RegisteredSecretUpdateCallback(path string, callback func(path string, secrets map[string]string))
```

bnevis-i: Seems to be that if we have a Register() that we should have an UnRegister() as well.

bootstrap/secret/secure.go
```diff
@@ -177,6 +178,9 @@ func (p *SecureProvider) StoreSecret(path string, secrets map[string]string) err
 		return err
 	}
 
+	// Execute Callbacks on registered secret paths.
+	p.SecretsUpdatedWithPath(path)
+
```
bnevis-i, line 183: See my earlier point about not exposing this via the public interface and supplying the new secret value since we have it anyway.

bnevis-i (commented):
I am thinking that SecretsUpdated() and SecretsUpdatedWithPath() should have never been public interface methods to begin, that the callback should send a copy of the current secret value, and that the Register() method should return a function object to unregister the callback that was added.

bnevis-i:
@drkfmorton Please update PR title.

bootstrap/config/config.go
```diff
@@ -511,7 +512,11 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
 					lc.Info("Insecure Secrets have been updated")
 					secretProvider := container.SecretProviderFrom(cp.dic.Get)
 					if secretProvider != nil {
-						secretProvider.SecretsUpdated()
+						// Find the updated secret's path and perform call backs.
+						updatedSecrets := getMapDiff(previousInsecureSecrets, currentInsecureSecrets)
+						for _, v := range updatedSecrets {
+							secretProvider.SecretsUpdatedWithPath(v)
+						}
```
jim-wang-intel, line 519: Should we make sure this update is protected by mutex to have threading safe?

bootstrap/secret/secure.go
```diff
@@ -388,3 +392,25 @@ func (p *SecureProvider) ListSecretPaths() ([]string, error) {
 
 	return secureSecrets, nil
 }
+
+// RegisteredSecretUpdateCallback registers a callback for a secret.
+func (p *SecureProvider) RegisteredSecretUpdateCallback(path string, callback func(path string)) {
```
jim-wang-intel, line 397: do we have error on callback? if so, the callback func should return error?

lenny-intel:
> I am thinking that SecretsUpdated() and SecretsUpdatedWithPath() should have never been public interface methods to begin, that the callback should send a copy of the current secret value, and that the Register() method should return a function object to unregister the callback that was added.

SecretsUpdated() had to be as it is called from outside this package when the InsecureSecrets are updated from Consul.

bootstrap/config/config.go
bnevis-i: Agree.  We don't want to diff the map if someone is in the middle of updating it.

bnevis-i:
> > I am thinking that SecretsUpdated() and SecretsUpdatedWithPath() should have never been public interface methods to begin, that the callback should send a copy of the current secret value, and that the Register() method should return a function object to unregister the callback that was added.
> 
> SecretsUpdated() had to be as it is called from outside this package when the InsecureSecrets are updated from Consul.

@drkfmorton 

Does our callback trigger when SecretsUpdated() is called?  Does it need to?  If so, this might limit what we can do in the callback (e.g. not provide the path that was updated nor the value of the new secret?)

lenny-intel:
> Remove SecretsUpdatedWithPath() from the public interface.

@bnevis-i , it has been a few weeks since @drkfmorton and I discussed the approach for this, but I think we decided that `SecretsUpdatedWithPath()` would replace use of `SecretsUpdated()` so that we would have the path. This requires a change to the processing when we determine `InsecureSecrets` section has changed. i.e. determine which secret changed and call `SecretsUpdatedWithPath()` rather than `SecretsUpdated()`.

bnevis-i:
> > Remove SecretsUpdatedWithPath() from the public interface.
> 
> @bnevis-i , it has been a few weeks since @drkfmorton and I discussed the approach for this, but I think we decided that `SecretsUpdatedWithPath()` would replace use of `SecretsUpdated()` so that we would have the path. This requires a change to the processing when we determine `InsecureSecrets` section has changed. i.e. determine which secret changed and call `SecretsUpdatedWithPath()` rather than `SecretsUpdated()`.

Ok.  Keep both and we will refactor the old one away.

bootstrap/config/config.go
```diff
@@ -550,3 +555,19 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
 func (cp *Processor) logConfigInfo(message string, overrideCount int) {
 	cp.lc.Infof("%s (%d envVars overrides applied)", message, overrideCount)
 }
+
+func getMapDiff(m1 config.InsecureSecrets, m2 config.InsecureSecrets) []string {
```
lenny-intel, line 559:
```suggestion
func getSecretPathsChanged(m1 config.InsecureSecrets, m2 config.InsecureSecrets) []string {
```

bootstrap/config/config.go
```diff
@@ -550,3 +555,19 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
...
+	var updatedPaths []string
```
lenny-intel, line 560: Also needs to handle newly add Paths and probably Paths that got removed/changed

bootstrap/config/config.go
```diff
@@ -511,7 +512,11 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
 					lc.Info("Insecure Secrets have been updated")
 					secretProvider := container.SecretProviderFrom(cp.dic.Get)
 					if secretProvider != nil {
-						secretProvider.SecretsUpdated()
```
lenny-intel, line 514: Since this is internal code and should be the only place SecretsUpdated() is called, we may be able to remove it.

bootstrap/interfaces/secret.go
lenny-intel, line 33:
```suggestion
	// RegisteredSecretUpdatedCallback registers a callback for a secret.
	RegisteredSecretUpdatedCallback(path string, callback func(path string))
```

bootstrap/interfaces/secret.go
lenny-intel: Per my other comments `SecretUpdatedAtPath` will need to be called from the config processing outside this package so must be public. And it replaces `SecretsUpdated` which we may be able to remove.

bootstrap/secret/insecure.go
```diff
@@ -146,3 +147,25 @@ func (p *InsecureProvider) ListSecretPaths() ([]string, error) {
...
+		// Execute Callback for provided path.
+		for k, v := range p.registeredSecretCallbacks {
+			if k == path {
+				v(path)
+				return
+			}
+		}
+	}
+
+	p.lc.Infof("no callback registered for path: '%s'", path)
```
lenny-intel, line 170:
```suggestion
	p.lc.Debugf("no Secret Updated callback registered for secret path: '%s'", path)
```

bootstrap/secret/secure.go
lenny-intel: Per other comments this sadly needs to be public. I.e. this interface wasn't originally design to be used directly by the custom App and Device services. We'll fix it in 3.0.

bootstrap/secret/secure.go
```diff
@@ -388,3 +392,25 @@ func (p *SecureProvider) ListSecretPaths() ([]string, error) {
...
+		// Execute Callback for provided path.
+		for k, v := range p.registeredSecretCallbacks {
+			if k == path {
+				v(path)
+				return
+			}
+		}
+	}
+
+	p.lc.Infof("no callback registered for path: '%s'", path)
```
lenny-intel, line 415:
```suggestion
	p.lc.Debugf("no Secret Updated callback registered for secret path: '%s'", path)
```

lenny-intel (changes requested):

bootstrap/interfaces/secret.go
bnevis-i: Actually, "RegisterSecretUpdatedCallback" not "Register**ED**SecretUpdateCallback"

drkfmorton:
> > > I am thinking that SecretsUpdated() and SecretsUpdatedWithPath() should have never been public interface methods to begin, that the callback should send a copy of the current secret value, and that the Register() method should return a function object to unregister the callback that was added.
> > 
> > 
> > SecretsUpdated() had to be as it is called from outside this package when the InsecureSecrets are updated from Consul.
> 
> @drkfmorton
> 
> Does our callback trigger when SecretsUpdated() is called? Does it need to? If so, this might limit what we can do in the callback (e.g. not provide the path that was updated nor the value of the new secret?)
> 
> So will scale back my request. Remove SecretsUpdatedWithPath() from the public interface.

Currently, callbacks are not triggered when SecretsUpdated() is called, but when SecretsUpdatedWithPath is called. Otherwise, if SecretsUpdatedWithPath isn't called from any methods that are triggered from a secret is updated then callbacks will not be triggered.

bnevis-i:
> Currently, callbacks are not triggered when SecretsUpdated() is called, but when SecretsUpdatedWithPath is called. Otherwise, if SecretsUpdatedWithPath isn't called from any methods that are triggered from a secret is updated then callbacks will not be triggered.

That's ok.  Per discussion above we'll refactor SecretsUpdated() away once this PR is in.

drkfmorton:
recheck

bootstrap/interfaces/secret.go
```diff
@@ -28,4 +28,10 @@ type SecretProvider interface {
 
 	// HasSecret returns true if the service's SecretStore contains a secret at the specified path.
 	HasSecret(path string) (bool, error)
+
+	// RegisteredSecretUpdatedCallback registers a callback for a secret.
+	RegisteredSecretUpdatedCallback(path string, callback func(path string))
```
bnevis-i, line 33: Make this function return `(func(), error)`.  When called, the returned func() will remove the registered callback from the map (only if path and callback match).

bootstrap/secret/insecure.go
```diff
@@ -146,3 +148,24 @@ func (p *InsecureProvider) ListSecretPaths() ([]string, error) {
 
 	return results, nil
 }
+
+// RegisteredSecretUpdatedCallback registers a callback for a secret.
+func (p *InsecureProvider) RegisteredSecretUpdatedCallback(path string, callback func(path string)) {
+	p.registeredSecretCallbacks[path] = callback
```
bnevis-i, line 154: This method as a subtle bug -- if path already has a callback, then this code will overwrite it.

Consequently, we either need to support multiple callbacks for a given path, or return an error if a callback is already registered.

As mentioned earlier, return a func() that if called will deregister the callback.

bootstrap/config/config.go
```diff
@@ -550,3 +556,34 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
...
+		} else {
+			// Handling for changes to a Path.
+			updatedPaths = append(updatedPaths, val1.Path, m2[key1].Path)
+		}
+	}
+
+	return cleanSecretPathArray(updatedPaths)
+}
+
+// cleanSecretPathArray cleans an array of secret paths.
```
bnevis-i, line 580: This comment is not descriptive enough as to why this is needed.  Why put a blank path in the map to begin with.  Could we have not checked this earlier instead of doing another pass on the array?

bootstrap/secret/secure.go
```diff
@@ -388,3 +393,24 @@ func (p *SecureProvider) ListSecretPaths() ([]string, error) {
...
+		// Execute Callback for provided path.
+		for k, v := range p.registeredSecretCallbacks {
+			if k == path {
+				v(path)
+				return
+			}
+		}
+	}
+
+	p.lc.Debugf("no callback registered for path: '%s'", path)
```
bnevis-i, line 415: Print the debug if we're calling the callback.  Don't need debugging for when we're not calling it.

bnevis-i (changes requested):

bootstrap/interfaces/secret.go
```diff
@@ -28,4 +28,12 @@ type SecretProvider interface {
...
 	// HasSecret returns true if the service's SecretStore contains a secret at the specified path.
 	HasSecret(path string) (bool, error)
+
+	// RegisteredSecretUpdatedCallback registers a callback for a secret.
+	RegisteredSecretUpdatedCallback(path string, callback func(path string)) (func(), error)
+
+	// SecretUpdatedAtPath performs updates and callbacks for an updated secret or path.
+	SecretUpdatedAtPath(path string)
+
+	DeregisterSecretUpdatedCallback(path string)
```
bnevis-i, line 38:
```suggestion
```

No Deregister method needed.

bootstrap/secret/insecure.go
```diff
@@ -146,3 +148,40 @@ func (p *InsecureProvider) ListSecretPaths() ([]string, error) {
...
+		}, fmt.Errorf("there is a callback already registered for path '%v'", path)
+	}
+
+	// Register new call back for path.
+	p.registeredSecretCallbacks[path] = callback
+
+	return func() {
+		p.DeregisterSecretUpdatedCallback(path)
+	}, nil
+}
```
bnevis-i, line 166:
```suggestion
func (p *InsecureProvider) RegisteredSecretUpdatedCallback(path string, callback func(path string)) (func(), error) {
	if _, ok := p.registeredSecretCallbacks[path]; ok {
		return nil, fmt.Errorf("there is a callback already registered for path '%s'", path)
	}

	// Register new call back for path.
	p.registeredSecretCallbacks[path] = callback

	return func() {
		if _, ok := p.registeredSecretCallbacks[path]; ok {
			delete(p.registeredSecretCallbacks, path)
		}
	}, nil
}
```

bnevis-i: 1. If there is an error, don't need to return the func()
2. Inline the code to deregister the function.  Don't need a public Deregister.

bootstrap/secret/insecure.go
```diff
@@ -146,3 +148,40 @@ func (p *InsecureProvider) ListSecretPaths() ([]string, error) {
...
+
+// SecretUpdatedAtPath performs updates and callbacks for an updated secret or path.
+func (p *InsecureProvider) SecretUpdatedAtPath(path string) {
+	p.lastUpdated = time.Now()
+	if p.registeredSecretCallbacks != nil {
+		// Execute Callback for provided path.
+		for k, v := range p.registeredSecretCallbacks {
+			if k == path {
+				v(path)
+				p.lc.Debugf("callback registered for path: '%s'", path)
```
bnevis-i, line 176:
```suggestion
				p.lc.Debugf("invoking callback registered for path: '%s'", path)
				v(path)
```


We need to know what the code was doing BEFORE it calls the callback

bootstrap/secret/secure.go
```diff
@@ -388,3 +393,42 @@ func (p *SecureProvider) ListSecretPaths() ([]string, error) {
...
+}
+
+func (p *SecureProvider) DeregisterSecretUpdatedCallback(path string) {
+	if _, ok := p.registeredSecretCallbacks[path]; ok {
+		delete(p.registeredSecretCallbacks, path)
+	}
+}
```
bnevis-i, line 434:
```suggestion
```

bnevis-i (changes requested):

bnevis-i:
@drkfmorton Please run "make test" before submitting next PR and fix any issues found.

bnevis-i (approved):

bootstrap/config/config.go
```diff
@@ -550,3 +555,29 @@ func (cp *Processor) listenForChanges(serviceConfig interfaces.Configuration, co
...
+			}
+
+			if len(m2[key1].Path) > 0 {
+				updatedPaths = append(updatedPaths, m2[key1].Path)
+			}
+		}
+	}
+
+	return updatedPaths
+}
```
lenny-intel, line 583: This need to be refactored to do deep compare of entries in the top level map if neither are nil. Something like this

```suggestion
func getSecretPathsChanged(prevVals config.InsecureSecrets, curVals config.InsecureSecrets) []string {
	var updatedPaths []string
	
       for key, prevVal := range prevVals{
          curVal:= curVals[key]
          // Catches removed secrets
          if curVal== nil {
             updatedPaths = append(updatedPaths, prevVal.Path)
             continue
          }

         // Catches changes to secret data or to the path name
         if !reflect.DeepEqual(prevVal,curVal]) {
              updatedPaths = append(updatedPaths, curVal.Path)

             // Catches path name changes, so also include the previous path
             if prevVal.path != curVal.path {
                     updatedPaths = append(updatedPaths, prevVal.path)
             }
         }
      }
      
      for key, curVal := range curVals{
         // Catches new secrets added
         if prevVals[key] == nil {
            updatedPaths = append(updatedPaths, curVal.Path)
          }
       }

	return updatedPaths
}
```

bootstrap/handlers/httpserver.go
```diff
@@ -116,8 +116,9 @@ func (b *HttpServer) BootstrapHandler(
 	}).HandlerFunc(HandlePreflight(bootstrapConfig.Service.CORSConfiguration))
 
 	server := &http.Server{
-		Addr:    addr,
-		Handler: b.router,
+		Addr:              addr,
+		Handler:           b.router,
+		ReadHeaderTimeout: 5 * time.Second, // G112: A configured ReadHeaderTimeout in the http.Server averts a potential Slowloris Attack
```
lenny-intel, line 121: Why is this part of this PR. Are you using newer version of the linter? 
@bnevis-i , are we ok with hard coded 5 secs for this?

lenny-intel (changes requested):
Still more to review, but wanted to get these comments posted.

bootstrap/handlers/httpserver.go
bnevis-i: @lenny-intel Yes.  The number of seconds is somewhat arbitrary.  The important thing is to have a limit.

bnevis-i (approved):

bootstrap/config/config_test.go
```diff
@@ -0,0 +1,98 @@
...
+				Secrets: map[string]string{
+					UsernameKey: expectedUsername,
+					PasswordKey: expectedPassword,
+				}}}},
+		{"Valid - No updates, unsorted keys", nil, curVals, config.InsecureSecrets{
+			"DB": config.InsecureSecretsInfo{
+				Path: expectedPath,
+				Secrets: map[string]string{
+					UsernameKey: expectedUsername,
+					PasswordKey: expectedPassword,
```
lenny-intel, line 88: This order is the same as in curVals, switch and verify test still passes.
```suggestion
					PasswordKey: expectedPassword,
					UsernameKey: expectedUsername,
```

bootstrap/config/config_test.go
```diff
@@ -0,0 +1,98 @@
...
+				}}}},
+		{"Valid - Path delete", []string{expectedPath, ""}, prevVals, config.InsecureSecrets{
+			"DB": config.InsecureSecretsInfo{
+				Path: "",
+				Secrets: map[string]string{
+					UsernameKey: expectedUsername,
+					PasswordKey: expectedPassword,
+				}}}},
```
lenny-intel, line 82: Here it should be as the whole entry was removed.
```suggestion
		{"Valid - Path delete", []string{expectedPath, ""}, prevVals, config.InsecureSecrets{},
```

bootstrap/config/config_test.go
```diff
@@ -0,0 +1,98 @@
...
+		{"Valid - Deleted Secret", []string{expectedPath}, prevVals, config.InsecureSecrets{
+			"DB": config.InsecureSecretsInfo{
+				Path: expectedPath,
+				Secrets: map[string]string{
+					UsernameKey: expectedUsername,
+				}}}},
+		{"Valid - Path update", []string{expectedPath}, prevVals, config.InsecureSecrets{
+			"DB": config.InsecureSecretsInfo{
+				Path: "redisdb",
```
lenny-intel, line 71: This is the same as `expectedPath` so how is the `Path update`?
Also is it the new path or the old path we expect to be returned or both?
```suggestion
				Path: "message-bus",
```

bootstrap/secret/insecure_test.go
```diff
@@ -192,3 +192,103 @@ func TestInsecureProvider_HasSecrets(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.Name, func(t *testing.T) {
+			target := NewInsecureProvider(tc.Config, logger.MockLogger{})
+
+			if tc.Callback != nil {
+				target.registeredSecretCallbacks[tc.Path] = tc.Callback
+			}
+
+			target.SecretUpdatedAtPath(tc.Path)
+		})
```
lenny-intel, line 225: Missing test validation?

bootstrap/secret/secure_test.go
```diff
@@ -419,3 +419,89 @@ func TestSecureProvider_ListSecretPathsSecrets(t *testing.T) {
...
+		t.Run(tc.Name, func(t *testing.T) {
+			target := NewSecureProvider(context.Background(), tc.Config, logger.MockLogger{}, nil, nil, "testService")
+			target.SetClient(tc.Client)
+
+			if tc.Callback != nil {
+				target.registeredSecretCallbacks[tc.Path] = tc.Callback
+			}
+
+			target.SecretUpdatedAtPath(tc.Path)
+		})
```
lenny-intel, line 446: Missing test validation?

bootstrap/secret/secure_test.go
```diff
@@ -419,3 +419,89 @@ func TestSecureProvider_ListSecretPathsSecrets(t *testing.T) {
...
+		Callback func(path string)
+	}{
+		{"Valid Secure", TestConfig{}, mock, expectedPath, func(path string) {}},
+		{"Valid No Callbacks", TestConfig{}, mock, expectedPath, nil},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.Name, func(t *testing.T) {
+			target := NewSecureProvider(context.Background(), tc.Config, logger.MockLogger{}, nil, nil, "testService")
+			target.SetClient(tc.Client)
```
lenny-intel, line 467: Is the mock really needed since you don't have an mock.On() calls? i.e. the client is never called by this test.

bootstrap/secret/secure_test.go
```diff
@@ -419,3 +419,89 @@ func TestSecureProvider_ListSecretPathsSecrets(t *testing.T) {
...
+				assert.NotEmpty(t, target.registeredSecretCallbacks[tc.Path])
+			} else {
+				assert.Nil(t, target.registeredSecretCallbacks[tc.Path])
+			}
+		})
+	}
+}
+
+func TestSecureProvider_DeregisterSecretUpdatedCallback(t *testing.T) {
+	mock := &mocks.SecretClient{}
```
lenny-intel, line 481: Same not need for mock.

lenny-intel (changes requested):

lenny-intel (approved):
LGTM

bootstrap/config/config_test.go
```diff
@@ -0,0 +1,94 @@
+/*******************************************************************************
+ * Copyright 2020 Intel Corp.
```
ajcasagrande, line 2: Should this be 2022?

lenny-intel:
Looks like the order of paths when path name changes can variy

For failed test in Pipeline:
```
[2022-09-29T21:32:02.364Z] --- FAIL: TestInsecureProvider_ListPaths (0.00s)

[2022-09-29T21:32:02.364Z]     --- FAIL: TestInsecureProvider_ListPaths/Valid (0.00s)

[2022-09-29T21:32:02.364Z]         insecure_test.go:141: 

[2022-09-29T21:32:02.364Z]             	Error Trace:	/w/workspace/xfoundry_go-mod-bootstrap_PR-373/bootstrap/secret/insecure_test.go:141

[2022-09-29T21:32:02.364Z]             	Error:      	Not equal: 

[2022-09-29T21:32:02.364Z]             	            	expected: []string{"redisdb", "kongdb"}

[2022-09-29T21:32:02.364Z]             	            	actual  : []string{"kongdb", "redisdb"}

[2022-09-29T21:32:02.364Z]             	            	

[2022-09-29T21:32:02.364Z]             	            	Diff:

[2022-09-29T21:32:02.364Z]             	            	--- Expected

[2022-09-29T21:32:02.364Z]             	            	+++ Actual

[2022-09-29T21:32:02.364Z]             	            	@@ -1,4 +1,4 @@

[2022-09-29T21:32:02.364Z]             	            	 ([]string) (len=2) {

[2022-09-29T21:32:02.364Z]             	            	- (string) (len=7) "redisdb",

[2022-09-29T21:32:02.364Z]             	            	- (string) (len=6) "kongdb"

[2022-09-29T21:32:02.364Z]             	            	+ (string) (len=6) "kongdb",

[2022-09-29T21:32:02.364Z]             	            	+ (string) (len=7) "redisdb"

[2022-09-29T21:32:02.364Z]             	            	 }

[2022-09-29T21:32:02.364Z]             	Test:       	TestInsecureProvider_ListPaths/Valid

[2022-09-29T21:32:02.364Z] FAIL

```

lenny-intel (approved):
LGTM

PR #502, opened 2023-03-29
fix: Rework merge of configs to better handle empty values

### LINKED ISSUE #501: [WIP BUG] Empty value not handle properly when merging Common & Private config
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: Config

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

### Description and Minimal Reproduction [**REQUIRED**]

Settings with empty values are removed when pulled from Config Provider.
Need better way to remove unspecified empty values from configure p[prior to merging into main config.

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:**

**Anything else relevant?**

lenny-intel:
fixes #501

## PR Checklist

  <link to docs PR>

## Testing Instructions
Build Core Data/ App Template/ Device Simple with this branch
Run each and verify the following
- Configs are loaded from file and pushed to Config Provider as expected
-   - Common configs and private configs merged without losing empty setting in private config
- Configs are loaded from Config Provider as expected
   - Common configs and private configs merged without losing empty setting in private config
- Writable changes are handled at all levels 
    - Private
    - All Common
    - App or Device common

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -48,11 +48,10 @@ import (
 )
 
 const (
-	writableKey       = "/Writable"
+	writableKey       = "Writable"
 	allServicesKey    = "all-services"
 	appServicesKey    = "app-services"
 	deviceServicesKey = "device-services"
-	pathSep           = "/"
 )
 
 // UpdatedStream defines the stream type that is notified by ListenForChanges when a configuration update is received.
@@ -169,7 +168,20 @@ func (cp *Processor) Process(
 			if err := cp.loadConfigFromProvider(privateServiceConfig, privateConfigClient); err != nil {
 				return err
 			}
-			if err := utils.MergeValues(serviceConfig, privateServiceConfig); err != nil {
+			configKeys, err := privateConfigClient.GetConfigurationKeys("")
+			if err != nil {
+				return err
+			}
+
+			// Must remove any settings in the config that are not actually present in the Config Provider
+			privateConfigKeys := utils.StringSliceToMap(configKeys)
+			privateConfigMap, err := utils.RemoveUnusedSettings(privateServiceConfig, utils.BuildBaseKey(configStem, serviceKey), privateConfigKeys)
+			if err != nil {
+				return fmt.Errorf("could not remove unused settings from private configurations: %s", err.Error())
+			}
+
+			// Now merge only the actual present value with the existing configuration from common.
+			if err := utils.MergeValues(serviceConfig, privateConfigMap); err != nil {
 				return fmt.Errorf("could not merge common and private configurations: %s", err.Error())
 			}
 
@@ -208,7 +220,7 @@ func (cp *Processor) Process(
 		}
 		cp.lc.Infof("Private configuration loaded from file with %d overrides applied", overrideCount)
 
-		if err := cp.mergeMapWithConfig(serviceConfig, configMap); err != nil {
+		if err := utils.MergeValues(serviceConfig, configMap); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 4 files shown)


bootstrap/config/config.go
```diff
@@ -48,11 +48,10 @@ import (
 )
 
 const (
-	writableKey       = "/Writable"
+	writableKey       = "Writable"
```
ejlee3, line 51: are there any changes where we needed writableKey to have the `/` and it doesn't because of this change?

bootstrap/utils/utils.go
```diff
@@ -68,45 +70,62 @@ func MergeMaps(dest map[string]any, src map[string]any) {
...
+		return nil, fmt.Errorf("could not create map from %T: %s", src, err.Error())
+	}
+
+	removeMapUnusedSettings(srcMap, baseKey, usedSettingKeys)
+
+	return srcMap, nil
+}
+
+// removeMapUnusedSettings iterates over a map and removes any settings not in list of valid keys
+func removeMapUnusedSettings(target map[string]any, baseKey string, validKeys map[string]any) {
```
ejlee3, line 86: does this name fit better?
```suggestion
func removeUnusedSettingsFromMap(target map[string]any, baseKey string, validKeys map[string]any) {
```

bootstrap/config/config.go
```diff
@@ -577,36 +611,20 @@ func (cp *Processor) loadConfigYamlFromFile(yamlFile string) (map[string]any, er
...
-}
-
 // GetConfigFileLocation uses the environment variables and flags to determine the location of the configuration
 func GetConfigFileLocation(lc logger.LoggingClient, flags flags.Common) string {
 	configDir := environment.GetConfigDir(lc, flags.ConfigDirectory())
 	profileDir := environment.GetProfileDir(lc, flags.Profile())
 	configFileName := environment.GetConfigFileName(lc, flags.ConfigFileName())
 
-	return configDir + pathSep + profileDir + configFileName
+	return configDir + utils.PathSep + profileDir + configFileName
```
ejlee3, line 620: should this use a filepathpath.Join? that way it is os independent? also does profileDir already have a trailing PathSep?

bootstrap/config/config.go
lenny-intel: Nope, looked at the code that this is passed to and it looks for the `/` and strips it off if it has it.

bootstrap/config/config.go
lenny-intel: Nope, because this is a Consul path which is always Linux style.

lenny-intel: Yes, Good catch!

bootstrap/utils/utils.go
lenny-intel: done

bootstrap/config/config.go
lenny-intel: reworked

bootstrap/utils/utils_test.go
```diff
@@ -123,32 +152,82 @@ func TestMergeConfigs(t *testing.T) {
...
 	assert.NotZero(t, serviceConfig.Writable.StoreAndForward.MaxRetryCount)
 	assert.NotEmpty(t, serviceConfig.Trigger.Type)
 }
 
-func TestRemoveZeroValues(t *testing.T) {
-	config := ConfigurationMockStruct{
-		Registry: config.RegistryInfo{
-			Host: "localhost",
-			Port: 8500,
+func TestRemoveUnusedSettings(t *testing.T) {
```
ejlee3, line 160: Is this supposed to be changed?

```suggestion
func TestRemoveUnusedSettingsFromMap(t *testing.T) {
```

lenny-intel: TestRemoveUnusedSettings is correct

ejlee3 (approved):
LGTM!

jim-wang-intel (approved):

PR #508, opened 2023-04-11
refactor!: Rework SecretProvider interface so App/Device Services have limited API

### LINKED ISSUE #398: Create limited Secret Provider interface for App and Device services
Current Secret Provider interface exposes APIs meant for internal use only. Need to split out the internal API so service only see API appropriate for them to use. Naming should stay the same for version services use and the internal API will have the new name.

The full extended renamed interface should used by Bootstrap. That way the Device and App services' code doesn't have to change. Something like `SecretProviderExt` and have it inherit from `SecretProvider` interface (limited API).

lenny-intel:
BREAKING CHANGE: Service that need full SecretProvider API now use SecretProviderExt. Extra APIs have been removed for App/Device Services.

closes #398

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
 https://github.com/edgexfoundry/edgex-docs/pull/1020

## Testing Instructions
See https://github.com/edgexfoundry/edgex-go/pull/4510 testing

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -111,7 +111,7 @@ func RunAndReturnWaitGroup(
 
 	envVars := environment.NewVariables(lc)
 
-	var secretProvider interfaces.SecretProvider
+	var secretProvider interfaces.SecretProviderExt
 	if useSecretProvider {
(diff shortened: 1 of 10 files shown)


jim-wang-intel (approved):
LGTM

PR #525, opened 2023-05-01
fix: InsecureSecrets change detection

### LINKED ISSUE #524: InsecureSecrets change detection does not work
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: 

### Is this a regression?

No

### Description and Minimal Reproduction [**REQUIRED**]

When InsecureSecrets are updated, the code that runs when a value is changed never gets called. This is because InsecureSecrets is a map data-type and the fields are not clones, so what happens is the original data gets clobbered when `utils.MergeValues` is called. This leads to `reflect.DeepEqual(currentInsecureSecrets, previousInsecureSecrets)` always being true because they point to the same data internally.

https://github.com/edgexfoundry/go-mod-bootstrap/blob/4ba5682ebe8da857aa6bdd457b4c74b30b68e3b1/bootstrap/config/config.go#L834-L842

https://github.com/edgexfoundry/go-mod-bootstrap/blob/4ba5682ebe8da857aa6bdd457b4c74b30b68e3b1/bootstrap/config/config.go#L856-L858

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:**

**Anything else relevant?**

I have confirmed this to be the case through real-time debugging and seeing the value of `previousInsecureSecrets` change between the time it is set and the time it is checked.

ajcasagrande:
fixes #524

## PR Checklist

- [n/a?] I have opened a PR for the related docs change (if not, why?)
 N/A due to no functionality change

## Testing Instructions

- Run EdgeX in non-secure mode
- Run a service with a secret callback code
- Change an InsecureSecret
- Verify that the callback was triggered

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -831,9 +831,13 @@ func (cp *Processor) isPrivateOverride(previous any, updated any, privateConfigC
 
 func (cp *Processor) applyWritableUpdates(serviceConfig interfaces.Configuration, raw any) {
 	lc := cp.lc
-	previousInsecureSecrets := serviceConfig.GetInsecureSecrets()
 	previousLogLevel := serviceConfig.GetLogLevel()
 	previousTelemetryInterval := serviceConfig.GetTelemetryInfo().Interval
+
+	var previousInsecureSecrets config.InsecureSecrets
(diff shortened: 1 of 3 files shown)


ejlee3 (approved):
LGTM

lenny-intel (commented):
LGTM

lenny-intel (approved):

Issue #550, opened 2023-05-31
External MQTT Rename SecretPath to SecretName


FelixTing:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: Config

### Is this a regression?

 Yes, the previous version in which this bug was not present was: Levski

### Description and Minimal Reproduction [**REQUIRED**]

- Run EdgeX 3.x in non-secure mode
- Modify the configuration of edgex-core-command via Consul to enable ExternalMQTT with the `usernamepassword` authentication mode:
  - Set ExternalMQTT.Enabled to `true`
  - Set ExternalMQTT.AuthMode to `usernamepassword`
- Restart edgex-core-command

## 🔥 Exception or Error
<pre><code>
level=ERROR ts=2023-05-31T10:33:19.103500932Z app=core-command source=externalmqtt.go:77 msg="Failed to retrieve secret data: Error, secretName () doesn't exist in secret store"
</code></pre>

## 🌍  Your Environment
**Deployment Environment:** Docker

**EdgeX Version [**REQUIRED**]:** main 3.x

**Anything else relevant?**

cloudxxx8:
@cherrycl please add related test for this issue in the next release

PR #551, opened 2023-05-31
fix: Rename SecretPath to SecretName

FelixTing:
fix: #550 

## PR Checklist

  https://github.com/edgexfoundry/edgex-docs/pull/1095

### DIFF
--- a/bootstrap/handlers/externalmqtt.go
+++ b/bootstrap/handlers/externalmqtt.go
@@ -72,14 +72,14 @@ func (e *ExternalMQTT) BootstrapHandler(ctx context.Context, wg *sync.WaitGroup,
 	}
 
 	//get the secrets from the secret provider and populate the struct
-	secretData, err := messaging.GetSecretData(authMode, brokerConfig.SecretPath, secretProvider)
+	secretData, err := messaging.GetSecretData(authMode, brokerConfig.SecretName, secretProvider)
 	if err != nil {
 		lc.Errorf("Failed to retrieve secret data: %s", err.Error())
 		return false
 	}
 	//ensure that the AuthMode selected has the required secret values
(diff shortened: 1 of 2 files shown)


cloudxxx8:
@lenny-intel please see whether we can fix this issue in 3.0

lenny-intel:
> @lenny-intel please see whether we can fix this issue in 3.0

@cloudxxx8 , Dang it. :-( We would have to delay the release for this and IMO it isn't worth it for a minor consistency issue. It will have to wait until 4.0.

cloudxxx8:
@lenny-intel it is a real bug, not a consistency issue, so we can fix it in v3.1.

lenny-intel:
> @lenny-intel it is a real bug, not a consistency issue, so we can fix it in v3.1.

@cloudxxx8 , How is this a real bug, rather than a name consistency issue? 

It is a breaking change to 3.0.0. Should discuss in Planning next week.

cloudxxx8:
@lenny-intel it's my misunderstanding of the error log.  I thought code tried to access `SecretName` anyway.
On the other hand it's the bug in the core-command config.  We are using `SecretName` in core-command.
https://github.com/edgexfoundry/edgex-go/blob/1fb6775d8804dda106fccace673c66bb187c3e57/cmd/core-command/res/configuration.yaml#L31

We have to change the config or merge this PR in v3.1.

lenny-intel:
@FelixTing , since the configuration YAML has the correct name, this is bug fix and not a breaking change as I first thought.
Please update the commit message an PR to remove specifying it as breaking change. THX!

lenny-intel (commented):
LGTM once specifying breaking change is removed.

lenny-intel (approved):
LGTM

PR #555, opened 2023-06-20
feat: Implement reusable load file function

### LINKED ISSUE #552: [URI4Files] Go- Implement reusable LoadFile function
See ADR here: https://docs.edgexfoundry.org/3.0/design/adr/0027-URIs%20for%20Files/

### Basic file loading
- Function defaults to standard local file load (`os.ReadFile`) if path doesn't start with `http://` or `https://`
- If path doesn't start with  `http://` or `https://` , then download file from the passed in URI. See [example here](https://golangcode.com/download-a-file-from-a-url/)

### Secure Credentials
- Implement URI query parameter `edgexSecretName` to specify name of the secret which contains the following
   - type (always set to  httpheader)
   - headername - Header key used used to add to HTTP Header
   - headercontents - Content to set in the header.
- Function will need access to the SecretProvider to Get the secret by name
- Credentials will be added to the HTTP Header prior to attempting to pull the file.

### Warning
The file URI may contain `username-password` (<username>:<password>@) in plain text. The LoadFile function should and callers should **NEVER** log the `username-password`  portion of the URI.

brian-intel:
close #552

## PR Checklist

  <link to docs PR>
Docs will be updated in future PR

## Testing Instructions

run "make test"


bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
 	return nil
 }
+
+func LoadFile(path string, contents *any) error {
+	var fileBytes []byte
+	var err error
+
+	lowerPath := strings.ToLower(path)
+
+	if strings.Contains(lowerPath, "http") {
```
lenny-intel, line 177: This needs to be a `Starts with` 
```suggestion
	if strings.Index(lowerPath, "http") == 0 {
```

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
 	}
 	return nil
 }
+
+func LoadFile(path string, contents *any) error {
```
lenny-intel, line 171: Must match signature of `os.ReadFile` which will be replaced with call to this function.

```suggestion
func LoadFile(path string) ([]byte, error) {
```

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
+			return fmt.Errorf("Could not unmarshal JSON (from %T) into type %T: %v", fileBytes, contents, err)
+		}
+	case ".yaml":
+		if err = yaml.Unmarshal(fileBytes, &contents); err != nil {
+			return fmt.Errorf("Could not unmarshal YAML (from %T) into type %T: %v", fileBytes, contents, err)
+		}
+	default:
+		return fmt.Errorf("Could not load unknown file type %s", pathExtension)
+	}
+	return nil
```
lenny-intel, line 212:
```suggestion
	return fileBytes, nil
```

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
+		if err = json.Unmarshal(fileBytes, &contents); err != nil {
+			return fmt.Errorf("Could not unmarshal JSON (from %T) into type %T: %v", fileBytes, contents, err)
+		}
+	case ".yaml":
+		if err = yaml.Unmarshal(fileBytes, &contents); err != nil {
+			return fmt.Errorf("Could not unmarshal YAML (from %T) into type %T: %v", fileBytes, contents, err)
+		}
+	default:
+		return fmt.Errorf("Could not load unknown file type %s", pathExtension)
+	}
```
lenny-intel, line 211: This goes beyond the scope of this function. Please remove.

bootstrap/utils/utils_test.go
```diff
@@ -278,3 +279,37 @@ func TestDeepCopy(t *testing.T) {
...
+		Contents    any
+		ExpectedErr string
+	}{
+		{"Valid - load from YAML file", path.Join("..", "config", "testdata", "configuration.yaml"), configuration, ""},
+		{"Valid - load from JSON file", path.Join(".", "testdata", "configuration.json"), configuration, ""},
+		{"Valid - load from HTTPS", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml", configuration, ""},
+		{"Invalid - File not found", "bogus", nil, "Could not read file"},
+		{"Invalid - load invalid file type", path.Join("..", "bootstrap.go"), nil, "Could not load unknown file type"},
+		// {"Invalid - load from YAML file", path.Join("..", "config", "testdata", "bogus.yaml"), configuration, "Could not unmarshal YAML"},
+		{"Invalid - load from JSON file", path.Join(".", "testdata", "bogus.json"), configuration, "Could not unmarshal JSON"},
```
lenny-intel, line 299: Beyond scope. Please remove

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
 }
+
+func LoadFile(path string, contents *any) error {
+	var fileBytes []byte
+	var err error
+
+	lowerPath := strings.ToLower(path)
+
+	if strings.Contains(lowerPath, "http") {
+		resp, err := http.Get(lowerPath)
```
lenny-intel, line 178: Will ned to use http.Request and request.Do in order to add the auth header and set an explicit timeout.

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
+		if err != nil {
+			return fmt.Errorf("Could not get remote file")
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode >= 400 {
```
lenny-intel, line 184: only the 200s are considered succes codes, but should we expect only 200 since the other 20x's don't apply?
https://umbraco.com/knowledge-base/http-status-codes/#2xx
```suggestion
		if resp.StatusCode >= 300 {
```

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
+			return fmt.Errorf("Invalid status code %d loading remote file", resp.StatusCode)
```
lenny-intel, line 185: Can we parse the URI and display the URI in message without the potential `username-password`?

bootstrap/utils/utils_test.go
```diff
@@ -278,3 +279,37 @@ func TestDeepCopy(t *testing.T) {
...
 }
+
+func TestLoadFile(t *testing.T) {
+	// Types
+	var configuration map[string]any
+
+	tests := []struct {
+		Name        string
+		Path        string
+		Contents    any
```
lenny-intel, line 290: This needs to be []byte

bootstrap/utils/utils_test.go
```diff
@@ -278,3 +279,37 @@ func TestDeepCopy(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.Name, func(t *testing.T) {
+			err := LoadFile(tc.Path, &tc.Contents)
+			if tc.ExpectedErr != "" {
+				assert.Contains(t, err.Error(), tc.ExpectedErr)
+				return
+			}
+
+			assert.NotNil(t, tc.Contents)
+			tc.Contents = make(map[string]any)
```
lenny-intel, line 312: Not sure what this is doing?

lenny-intel (changes requested):

bootstrap/utils/utils.go
ejlee3: how will we get the header and timeout information?

bootstrap/utils/utils.go
ejlee3: do we know how the username/password will be keyed off of - we weren't sure if this would be consistent across all users (pretty hard to guarantee).

bootstrap/utils/utils_test.go
ejlee3: this was resetting the configuration information - so that the variable can be reused - but this will likely get refactored out if we want to keep the signature of the LoadFile the same as ReadFile - just thought it would be nice to implement the marshaling here so that we can get rid of a lot of similar code.

bootstrap/utils/utils.go
lenny-intel: header is info is from the secret data from the SecretProvider as described in the ADR.
The timeout can be the `Service.RequestTimeout` from config except when loading the config file.
Best to add secretProvider and requestTimeout parameters to this function and leave it to callers to find the values to pass in..

bootstrap/utils/utils.go
lenny-intel: Yes, it is standard as outlined in the ADR. My guess is the URI parser will separate it out for you so you can get at the URL part to log.

bootstrap/utils/utils.go
```diff
@@ -161,3 +167,47 @@ func DeepCopy(src any, dest any) error {
...
 	return nil
 }
+
+func LoadFile(path string, contents *any) error {
+	var fileBytes []byte
+	var err error
+
+	lowerPath := strings.ToLower(path)
+
+	if strings.Index(lowerPath, "http") == 0 {
```
lenny-intel, line 177: I was thinking of a better way to ensure that the path is a URL and not a file name like `http-stuff.yaml` would be to use url.Parse. This now gives you the parsed URL to log the redacted url. 
```suggestion
	parsedUrl, err := urlParse(lowerPath, "http"
	if err == nil {
```

bootstrap/utils/utils.go
```diff
@@ -168,46 +167,41 @@ func DeepCopy(src any, dest any) error {
...
+		req, err := http.NewRequest("GET", path, nil)
+		// Set request header
+		//req.Header.Add("", "")
+		resp, err := client.Do(req)
 
-	if strings.Index(lowerPath, "http") == 0 {
-		resp, err := http.Get(lowerPath)
 		if err != nil {
-			return fmt.Errorf("Could not get remote file")
+			return nil, fmt.Errorf("Could not get remote file: %s", parsedUrl.Host)
```
lenny-intel, line 187:
```suggestion
			return nil, fmt.Errorf("Could not get remote file: %s", parsedUrl.Redacted())
```

bootstrap/utils/utils.go
```diff
@@ -168,46 +167,41 @@ func DeepCopy(src any, dest any) error {
...
 		}
 		defer resp.Body.Close()
 
-		if resp.StatusCode >= 400 {
-			return fmt.Errorf("Invalid status code %d loading remote file", resp.StatusCode)
+		if resp.StatusCode >= 300 {
+			return nil, fmt.Errorf("Invalid status code %d loading remote file: %s", resp.StatusCode, parsedUrl.Host)
```
lenny-intel, line 192:
```suggestion
			return nil, fmt.Errorf("Invalid status code %d loading remote file: %s", resp.StatusCode, parsedUrl.Redacted())
```

bootstrap/utils/utils.go
```diff
@@ -168,46 +167,41 @@ func DeepCopy(src any, dest any) error {
...
 		}
 
 		fileBytes, err = io.ReadAll(resp.Body)
 		if err != nil {
-			return fmt.Errorf("Could not read remote file: %v", err)
+			return nil, fmt.Errorf("Could not read remote file: : %s", parsedUrl.Host)
```
lenny-intel, line 197:
```suggestion
			return nil, fmt.Errorf("Could not read remote file: : %s", parsedUrl.Redacted())
```

lenny-intel (changes requested):

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+	var fileBytes []byte
+	var err error
+
+	client := &http.Client{
+		Timeout: timeout,
+	}
+
+	parsedUrl, err := url.Parse(path)
+	if parsedUrl.Scheme == "http" || parsedUrl.Scheme == "https" {
+		// resp, err := http.Get(path)
```
lenny-intel, line 24: remove commented out code

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/interfaces"
+)
+
+func LoadFile(path string, timeout time.Duration, provider interfaces.SecretProvider) ([]byte, error) {
+	var fileBytes []byte
+	var err error
+
+	client := &http.Client{
+		Timeout: timeout,
+	}
```
lenny-intel, line 20: Move down under condition for http request.

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+
+	parsedUrl, err := url.Parse(path)
+	if parsedUrl.Scheme == "http" || parsedUrl.Scheme == "https" {
+		// resp, err := http.Get(path)
+		req, err := http.NewRequest("GET", path, nil)
+
+		// Get httpheader secret
+		params := parsedUrl.Query()
+		if len(params) > 0 {
+			secrets, err := provider.GetSecret(params.Get("edgexSecretName"))
```
lenny-intel, line 30: Need to first verify `edgexSecretName` exists , i.e. there may be other query parameters

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			// Set request header
+			if len(secrets) > 0 {
```
lenny-intel, line 36: Also must verify the data contains the expected item, including `type` = `httpheader`

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+				req.Header.Add(secrets["headername"], secrets["headercontents"])
+			}
+		}
+
+		// Run request
+		resp, err := client.Do(req)
+
+		if err != nil {
+			return nil, fmt.Errorf("Could not get remote file: %s", parsedUrl.Host)
```
lenny-intel, line 45:
```suggestion
			return nil, fmt.Errorf("Could not get remote file: %s", parsedUrl.Redacted())
```

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode >= 300 {
+			return nil, fmt.Errorf("Invalid status code %d loading remote file: %s", resp.StatusCode, parsedUrl.Host)
```
lenny-intel, line 50:
```suggestion
			return nil, fmt.Errorf("Invalid status code %d loading remote file: %s", resp.StatusCode, parsedUrl.Redacted())
```

bootstrap/file/file.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		fileBytes, err = io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("Could not read remote file: : %s", parsedUrl.Host)
```
lenny-intel, line 55:
```suggestion
			return nil, fmt.Errorf("Could not read remote file: : %s", parsedUrl.Redacted())
```

lenny-intel (changes requested):

bootstrap/file/file_test.go
```diff
@@ -0,0 +1,58 @@
...
+		ContentLength      int
+		ExpectedErr        string
+		expectedSecretData map[string]string
+	}{
+		{"Valid - load from YAML file", path.Join("..", "config", "testdata", "configuration.yaml"), 4533, "", nil},
+		{"Valid - load from JSON file", path.Join(".", "testdata", "configuration.json"), 142, "", nil},
+		{"Valid - load from HTTPS", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml", 4533, "", nil},
+		{"Invalid - File not found", "bogus", 0, "Could not read file", nil},
+		{"Invalid - load from invalid HTTPS", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/configuration.yaml", 1, "Invalid status code", nil},
+		{"valid - load from HTTPS with secret", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml?edgexSecretName=mySecretName", 4533, "", map[string]string{"type": "httpheader", "headername": "Authorization", "headercontents": "Basic 1234567890"}},
```
ejlee3, line 30: Can we move this test case up to where the other valid ones are?
```suggestion
		{"Valid - load from HTTPS with secret", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml?edgexSecretName=mySecretName", 4533, "", map[string]string{"type": "httpheader", "headername": "Authorization", "headercontents": "Basic 1234567890"}},
```

bootstrap/file/file_test.go
```diff
@@ -0,0 +1,58 @@
...
+		{"invalid - load from HTTPS with invalid secret", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml?edgexSecretName=mySecretName", 4533, "", map[string]string{"type": "invalidheader", "headername": "Authorization", "headercontents": "Basic 1234567890"}},
```
vli11, line 31: "invalid" to "Invalid"

bootstrap/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+		if resp.StatusCode >= 300 {
+			return nil, fmt.Errorf("Invalid status code %d loading remote file: %s", resp.StatusCode, parsedUrl.Redacted())
+		}
+
+		fileBytes, err = io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("Could not read remote file: : %s", parsedUrl.Redacted())
+		}
+	} else {
+		fileBytes, err = os.ReadFile(path)
```
jim-wang-intel, line 57: we should also check if scheme is file:// right?  scheme from URL could be something else like `ftp`, `ws`, ... etc.

lenny-intel: No, per the ADR we are not supporting `file:// `. If not `http(s)` then use local file path.

bootstrap/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/interfaces"
+)
+
+func LoadFile(path string, timeout time.Duration, provider interfaces.SecretProvider) ([]byte, error) {
```
lenny-intel, line 14: Now that we moved this into `file` package, then function name should just be `Load` to avoid has in Go Lang is called name stuttering. (i.e. file.LoadFile)
```suggestion
func Load(path string, timeout time.Duration, provider interfaces.SecretProvider) ([]byte, error) {
```

bootstrap/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+	var fileBytes []byte
+	var err error
+
+	parsedUrl, err := url.Parse(path)
+	if parsedUrl.Scheme == "http" || parsedUrl.Scheme == "https" {
```
lenny-intel, line 19: err is never checked. It will be non-nil if local file path used. 
```suggestion
	if err != nil && (parsedUrl.Scheme == "http" || parsedUrl.Scheme == "https") {
```

bootstrap/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+		params := parsedUrl.Query()
+		edgexSecretName := params.Get("edgexSecretName")
+		if edgexSecretName != "" {
+			secrets, err := provider.GetSecret(edgexSecretName)
+			if err != nil {
+				return nil, err
+			}
+
+			// Set request header
+			if len(secrets) > 0 && secrets["type"] == "httpheader" {
```
lenny-intel, line 35: Should be an error if `type` != `httpheader`

bootstrap/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+				req.Header.Add(secrets["headername"], secrets["headercontents"])
```
lenny-intel, line 36: Must validate that `headername` and `headercontents` are actually present.

lenny-intel (changes requested):

bootstrap/file/file.go
jim-wang-intel: ok

bootstrap/file/file.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, fmt.Errorf("Could not parse file path: %v", err)
+	}
+
+	if parsedUrl.Scheme == "http" || parsedUrl.Scheme == "https" {
+		client := &http.Client{
+			Timeout: timeout,
+		}
+		req, err := http.NewRequest("GET", path, nil)
+		if err != nil {
+			return nil, fmt.Errorf("Unable to create new request: %v", err)
```
lenny-intel, line 29:
```suggestion
			return nil, fmt.Errorf("Unable to create new request for remote file: %s: %v", parsedUrl.Redacted(), err)
```

bootstrap/file/file.go
```diff
@@ -0,0 +1,77 @@
...
+			} else {
+				return nil, fmt.Errorf("Secret type is not httpheader")
+			}
+		}
+
+		// Run request
+		resp, err := client.Do(req)
+
+		if err != nil {
+			return nil, fmt.Errorf("Could not get remote file: %s", parsedUrl.Redacted())
```
lenny-intel, line 57:
```suggestion
			return nil, fmt.Errorf("Could not get remote file: %s: %v", parsedUrl.Redacted(), err)
```

bootstrap/file/file.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode >= 300 {
+			return nil, fmt.Errorf("Invalid status code %d loading remote file: %s", resp.StatusCode, parsedUrl.Redacted())
+		}
+
+		fileBytes, err = io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("Could not read remote file: : %s", parsedUrl.Redacted())
```
lenny-intel, line 67:
```suggestion
			return nil, fmt.Errorf("Could not read remote file: %s: %v", parsedUrl.Redacted(), err)
```

bootstrap/file/file_test.go
```diff
@@ -0,0 +1,58 @@
...
+		{"Valid - load from JSON file", path.Join(".", "testdata", "configuration.json"), 142, "", nil},
+		{"Valid - load from HTTPS", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml", 4533, "", nil},
+		{"Invalid - File not found", "bogus", 0, "Could not read file", nil},
+		{"Invalid - load from invalid HTTPS", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/configuration.yaml", 1, "Invalid status code", nil},
+		{"Valid - load from HTTPS with secret", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml?edgexSecretName=mySecretName", 4533, "", map[string]string{"type": "httpheader", "headername": "Authorization", "headercontents": "Basic 1234567890"}},
+		{"Invalid - load from HTTPS with invalid secret", "https://raw.githubusercontent.com/edgexfoundry/go-mod-bootstrap/main/bootstrap/config/testdata/configuration.yaml?edgexSecretName=mySecretName", 4533, "", map[string]string{"type": "invalidheader", "headername": "Authorization", "headercontents": "Basic 1234567890"}},
```
lenny-intel, line 31: Should have an `http` test case?

lenny-intel, line 31: ExpectedErr not set, so I would expect this test case to fail??

bootstrap/file/file_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	mockSecretProvider := &mocks.SecretProvider{}
+	mockSecretProvider.On("GetSecret", "").Return(nil)
+	mockSecretProvider.On("GetSecret", "notfound").Return(nil, errors.New("Not Found"))
```
lenny-intel, line 36: Where is this used? I don't see a `edgexSecretName=notfond` used above. I like the approach, but the test case is missing above.

lenny-intel (changes requested):

bootstrap/file/file_test.go
lenny-intel: This was not addressed, so unresolved the conversation

lenny-intel (approved):
LGTM

bootstrap/file/file.go
```diff
@@ -0,0 +1,77 @@
...
+				return nil, err
+			}
+
+			// Set request header
+			if len(secrets) > 0 && secrets["type"] == "httpheader" {
+				if secrets["headername"] != "" && secrets["headercontents"] != "" {
+					req.Header.Add(secrets["headername"], secrets["headercontents"])
+				} else {
+					return nil, fmt.Errorf("Secret headername and headercontents can not be empty")
+				}
```
jim-wang-intel, line 47: could simplify this but it's ok:
```suggestion
				if secrets["headername"] == "" || secrets["headercontents"] == "" {
				        return nil, fmt.Errorf("Secret headername and headercontents can not be empty")
				} 
					req.Header.Add(secrets["headername"], secrets["headercontents"])
```

jim-wang-intel (approved):

ejlee3 (approved):

Issue #564, opened 2023-07-18
Add better error handling in bootstrap handlers when Common Config is missing


lenny-intel:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for error handling in bootstrap handlers

### Description [**REQUIRED**]
 Currently if common config is not loaded we get an error like the following and the service crashes
```
source=httpserver.go:101 msg="unable to parse RequestTimeout value of  to a duration: time: invalid duration \"\""
```

### Describe the solution you'd like
 Each bootstrap handler must first check if the configuration it needs is populated. If not,  log error that the configuration is missing likely due to common config missing and return `false` to fail bootstrapping. 

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

PR #566, opened 2023-07-19
feat: add better error handlings when common config is missing

jim-wang-intel:
Gives a better error messages when the developer runs with hybrid mode but not including common config.

  Closes: #564

## PR Checklist

  <link to docs PR>

## Testing Instructions

- git clone this PR
- you can use device-simple from device-sdk repo to test this out when running in hybrid mode and don't provide common config (i.e. no -cc flag in place when running it)
- in the device sdk, change the go-mod-bootstrap import to the local cloned go-mod-bootstrap in go.mod file
- run `make` first from the base root of device-sdk repo to compile the device-sdk together with the local go-mod-bootstrap
- change directory to device-simple under example/cmd:  `cd ./example/cmd/device-simple`
- run make again to compile the device simple app: `make` 
- run with `./device-simple` without any flags like -cc or -d
- you should expect to see some error message like 
```console
"Service.RequestTime found empty in bootstrap conifg, missing common config? Use -cp or -cc flags for common config"
```
- you could now edit the configuration.yml in the res folder of device-simple to add the missing entries or items to fix this error message to go further until you see the app is starting or running without bootstrapping failed message in the log

### DIFF
--- a/bootstrap/handlers/clients.go
+++ b/bootstrap/handlers/clients.go
@@ -110,6 +110,11 @@ func (cb *ClientsBootstrap) BootstrapHandler(
 						return false
 					}
 
+					if len(config.GetBootstrap().Service.RequestTimeout) == 0 {
+						lc.Error("Service.RequestTimeout found empty in bootstrap config, missing common config? Use -cp or -cc flags for common config")
+						return false
+					}
+
 					// TODO: Move following outside loop when multiple messaging based clients exist
 					timeout, err := time.ParseDuration(config.GetBootstrap().Service.RequestTimeout)
 					if err != nil {
--- a/bootstrap/handlers/clients.go
+++ b/bootstrap/handlers/clients.go
@@ -111,7 +111,7 @@ func (cb *ClientsBootstrap) BootstrapHandler(
 					}
 
 					if len(config.GetBootstrap().Service.RequestTimeout) == 0 {
-						lc.Error("Service.RequestTimeout found empty in bootstrap config, missing common config? Use -cp or -cc flags for common config")
+						lc.Error("Service.RequestTimeout found empty in service's configuration, missing common config? Use -cp or -cc flags for common config")
 						return false
 					}
 
--- a/bootstrap/handlers/httpserver.go
+++ b/bootstrap/handlers/httpserver.go
@@ -89,7 +89,7 @@ func (b *HttpServer) BootstrapHandler(
 
 	if bootstrapConfig.Service.Port == 0 {
 		// should not be 0 as if it were set in local config
-		lc.Error("Service.Port is missing or should not be 0 in local private config")
+		lc.Error("Service.Port is missing from service's configuration or should not be 0 in local private config")
 		return false
 	}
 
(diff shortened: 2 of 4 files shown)


bootstrap/handlers/clients.go
```diff
@@ -110,6 +110,11 @@ func (cb *ClientsBootstrap) BootstrapHandler(
 						return false
 					}
 
+					if len(config.GetBootstrap().Service.RequestTimeout) == 0 {
+						lc.Error("Service.RequestTimeout found empty in bootstrap config, missing common config? Use -cp or -cc flags for common config")
```
lenny-intel, line 114:
```suggestion
						lc.Error("Service.RequestTimeout found empty in service's configuration, missing common config? Use -cp or -cc flags for common config")
```

bootstrap/handlers/httpserver.go
```diff
@@ -86,6 +87,12 @@ func (b *HttpServer) BootstrapHandler(
 
 	bootstrapConfig := container.ConfigurationFrom(dic.Get).GetBootstrap()
 
+	if bootstrapConfig.Service.Port == 0 {
+		// should not be 0 as if it were set in local config
+		lc.Error("Service.Port is missing or should not be 0 in local private config")
```
lenny-intel, line 92:
```suggestion
		lc.Error("Service.Port is missing from service's configuration or should not be 0 in local private config")
```

bootstrap/handlers/httpserver.go
```diff
@@ -96,6 +103,11 @@ func (b *HttpServer) BootstrapHandler(
 		addr = bootstrapConfig.Service.Host + ":" + port
 	}
 
+	if len(bootstrapConfig.Service.RequestTimeout) == 0 {
+		lc.Error("Service.RequestTimeout found empty in bootstrap config, missing common config? Use -cp or -cc flags for common config")
```
lenny-intel, line 107:
```suggestion
		lc.Error("Service.RequestTimeout found empty in service's configuration, missing common config? Use -cp or -cc flags for common config")
```

bootstrap/registration/registry.go
```diff
@@ -68,6 +72,10 @@ func createRegistryClient(
 		}
 	}
 
+	if len(bootstrapConfig.Registry.Host) == 0 || bootstrapConfig.Registry.Port == 0 || len(bootstrapConfig.Registry.Type) == 0 {
+		return nil, errors.New("Registry config is incomplete, missing common config? Use -cp or -cc flags for common config.")
+	}
```
lenny-intel, line 77: We don't need both checks. Since this one covers all missing, the other one can be removed.

lenny-intel (changes requested):

bootstrap/registration/registry.go
jim-wang-intel: ok, i'll update the error message slightly to include both cases.

bootstrap/handlers/clients.go
jim-wang-intel: all suggestions updated

lenny-intel (approved):
LGTM

PR #596, opened 2023-09-06
feat: Add new -rsi/--remoteServiceIPs flag and corresponding env override

### LINKED ISSUE #593: Add new -rs/--remoteService command line flag to set IP address required when service is running remotely
# 🚀 Feature Request

 This feature request is for `flags`

 When running a device or any other service remote from the core services (Core Metadata, Redis, etc) the IP address of the host where the core services must be set for the following:
 - Clients 
   - Core Metadata
   - Core Data
   - Core Command
   - etc
 - MessageBus
 - Database
 - Registry 

In addition the Service `Host` must be set to the remote host IP address for device service to receive commands from Core Command and the Service `ServerBindAddr` must be set so that in docker the host IP is not used the bind to for the WebServer.

 Managing all the above IP changes is complex and cumbersome and would be simplified by have a command line flag to takes the IP address of the local Host system and central Core service system and sets all values appropriately. 
Suggest the flag be called `-rs/--remoteService` which takes the IPs addresses in the following format: 
- `<local host IP>:<central host IP>:<server-bind IP>`
- Example: `--remoteService=192.0.1.8:192.0.1.21:0.0.0.0`

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

lenny-intel:
closes #593

## PR Checklist

  TBD

## Testing Instructions
TBD


bootstrap/config/config.go
```diff
@@ -283,9 +302,51 @@ func (cp *Processor) Process(
...
 	}
 
+	if remoteIPs != nil {
+		err = applyRemoteIPs(remoteIPs, serviceConfig)
+		if err != nil {
+			return err
+		}
+	}
+
 	return err
```
jim-wang-intel, line 312: this can just be simple as return call:
```suggestion
	return applyRemoteIPs(remoteIPs, serviceConfig)
```

lenny-intel: Nope, we don't want to call applyRemoteIPs if `remoteIPs ` is nil

jim-wang-intel: i got that, i guess you could move that check `remoteIPs` is nil into that func and just return nil as error in that case.

jim-wang-intel (approved):

lenny-intel:
@cloudxxx8 please review and let me know if you agree with the flag name. Doc PR is here: https://github.com/edgexfoundry/edgex-docs/pull/1213

bootstrap/container/dev_or_remote_mode.go
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+ * Copyright 2020 Intel Inc.
```
presatish, line 2:
```suggestion
 * Copyright 2023 Intel Inc.
```

lenny-intel: fixed

bootstrap/config/config.go
```diff
@@ -283,9 +303,51 @@ func (cp *Processor) Process(
 		}
 	}
 
+	if remoteHosts != nil {
```
jim-wang-intel, line 306: i am seeing an opportunity that you can absorb this `remoteHosts !=nil` check into your function `applyRemoteHosts`: so that you don't need to add those duplicate != nil check above and here.
The mental model is the same: if remoteHosts != nil means nothing to apply when that applyRemoteHosts() func is called.

jim-wang-intel: sorry i meant if remoteHosts == nil

lenny-intel: Moving this check of `remoteHosts` into `applyRemoteHosts()` func doesn't change the need to check `remoteHosts` in the code above. The check above under `useProvider` is specific to the host for the ConfigProvider (not in configuration) that has to be done before this call so the config is loaded for `applyRemoteHosts`

jim-wang-intel: ok, gotcha

bootstrap/container/dev_or_remote_mode.go
```diff
@@ -0,0 +1,36 @@
...
+
+type InDevOrRemoteMode struct {
+	Value bool
+}
+
+// DevOrRemoteModeName contains the name of the Dev or Remotes mode flag in the DIC.
+var DevOrRemoteModeName = di.TypeInstanceToName((*InDevOrRemoteMode)(nil))
+
+// DevOrRemoteModeFrom helper function queries the DIC and returns the Dev or Remotes mode flag.
+func DevOrRemoteModeFrom(get di.Get) bool {
```
cloudxxx8, line 29: Since it's the public di package, any future change may become a breaking change.
Do we always bundle Dev and Remote mode?  For the other service implementation, they might just need to confirm it's Dev mode or Remote mode.  I suggest to return a simple Service Mode struct contains `IsDev` and `IsRemote`, or implement two separated di container for `DevModeFrom` and `RemoteModeFrom`/`RemoteHostsFrom`.

cloudxxx8 (changes requested):

bootstrap/container/dev_or_remote_mode.go
lenny-intel: Good point. I like the structure approach. Will make that change.

jim-wang-intel (approved):

PR #659, opened 2024-02-01
feat: initial implemention of openziti to go-mod-bootstrap

### LINKED ISSUE #650: Add application embedded zero trust via openziti
# 🚀 Feature Request

bootstrap/handlers  This feature request is for...

 A clear and concise description of the problem or missing capability...

 If you have a solution in mind, please describe it.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

dovholuknf:
closes #650 

This is the initial PR to adopt OpenZiti into EdgexFoundry go-mod-bootstrap. Much of the functionality will be used in other repos. I've tried to be as thorough testing this as I could but I'm sure there will need to be work to fix/doc/etc. I wanted to get the code PR up as soon as possible to start getting feedback.

* Loggging - OpenZiti uses pfxlog which uses logrus. A hook is added to callback into EdgexFoundry's logger along with setting the default logger output to discard.
* auth_middleware is updated to allow connections from OpenZiti identities, defering authorization to OpenZiti
* SecretProvider now allows setting the HttpTransport to use. This is replaced by an OpenZiti implementation if zero trust is enabled, otherwise it's the default http transport
* any 'server' is now able to be removed from the IP-based underlay network entirely and able to be made available only via OpenZiti. This makes classic IP-based attacks impossible.
* SecurityOptions has been added to ServiceInfo and ClientInfo and is configured similarly to something like:
```
Service:
    SecurityOptions:
        Mode: "zerotrust"
        OpenZitiController: "openziti:1280"
        OpenZitiServiceName: "edgex.ui"
		
		
    core-command:
        Protocol: http
        Host: "core-command.edgex.ziti"
        Port: 80
        SecurityOptions:
            Mode: "zerotrust"
            OpenZitiController: "openziti:1280"
            OpenZitiServiceName: "edgex.core-command"
```

## PR Checklist

- [x ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [ ] I am not introducing a new dependency (add notes below if you are)
  <link to docs PR>

## Testing Instructions
This is a sweeping change that crosses numerous repos. I don't know how to effectively answer this particular question and this one change is insufficient for testing. A larger, cross-cutting PR/effort is needed to put all the pieces together. I could use help with knowing how to satisfy this 

## New Dependency Instructions (If applicable)
* github.com/openziti/sdk-golang (and transitives)

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -107,6 +107,7 @@ func RunAndReturnWaitGroup(
 		})
 	}
 
+	container.AdaptLogrusBasedLogging(dic)
 	translateInterruptToCancel(ctx, &wg, cancel)
 
 	envVars := environment.NewVariables(lc)
--- a/bootstrap/container/logging.go
+++ b/bootstrap/container/logging.go
@@ -15,7 +15,10 @@
 package container
 
 import (
+	"fmt"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/clients/logger"
+	"github.com/sirupsen/logrus"
+	"io"
 
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/di"
 )
@@ -31,4 +34,48 @@ func LoggingClientFrom(get di.Get) logger.LoggingClient {
 	}
 
 	return loggingClient
+}
+
+type LogrusAdaptor struct {
+	lc logger.LoggingClient
+}
+
+func (f *LogrusAdaptor) Format(entry *logrus.Entry) ([]byte, error) {
+	// Implement your custom formatting logic here
+	return []byte(fmt.Sprintf("[%s] %s\n", entry.Level, entry.Message)), nil
+}
+
+func (f *LogrusAdaptor) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+func (f *LogrusAdaptor) Fire(e *logrus.Entry) error {
+	switch e.Level {
+	case logrus.DebugLevel:
+		f.lc.Debug(e.Message)
+	case logrus.InfoLevel:
+		f.lc.Info(e.Message)
+	case logrus.WarnLevel:
+		f.lc.Warn(e.Message)
+	case logrus.ErrorLevel:
+		f.lc.Error(e.Message)
+	case logrus.FatalLevel:
+		f.lc.Error(e.Message)
+	case logrus.PanicLevel:
+		f.lc.Error(e.Message)
+	}
+
+	return nil
+}
+
+func AdaptLogrusBasedLogging(dic *di.Container) {
+	l := LoggingClientFrom(dic.Get)
+	// Create a new logger instance
+	hook := &LogrusAdaptor{
+		lc: l,
+	}
+	logrus.AddHook(hook)
+	logrus.SetFormatter(&logrus.TextFormatter{
+		DisableColors: true,
+	})
+	logrus.SetOutput(io.Discard)
 }
--- a/bootstrap/container/zerotrust.go
+++ b/bootstrap/container/zerotrust.go
@@ -1,0 +1,89 @@
+package container
+
+import (
+	"context"
+	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/interfaces"
+	"github.com/edgexfoundry/go-mod-bootstrap/v3/config"
+	"github.com/edgexfoundry/go-mod-core-contracts/v3/clients/logger"
+	edge_apis "github.com/openziti/sdk-golang/edge-apis"
+	"github.com/openziti/sdk-golang/ziti"
+	"net"
+	"net/http"
+	"strings"
+)
+
+const (
+	OpenZitiControllerKey = "OpenZitiController"
+	ZeroTrustKey          = "zerotrust"
+)
+
+func AuthToOpenZiti(ozController, jwt string) ziti.Context {
+	openZitiRootUrl := "https://" + ozController
+	if !strings.Contains(openZitiRootUrl, "://") {
+		openZitiRootUrl = "https://" + ozController
+	}
+	caPool, caErr := ziti.GetControllerWellKnownCaPool(openZitiRootUrl)
+	if caErr != nil {
+		panic(caErr)
+	}
+
+	credentials := edge_apis.NewJwtCredentials(jwt)
+	credentials.CaPool = caPool
+
+	cfg := &ziti.Config{
+		ZtAPI:       openZitiRootUrl + "/edge/client/v1",
+		Credentials: credentials,
+	}
+	cfg.ConfigTypes = append(cfg.ConfigTypes, "all")
+
+	ctx, ctxErr := ziti.NewContext(cfg)
+	if ctxErr != nil {
+		panic(ctxErr)
+	}
+	if err := ctx.Authenticate(); err != nil {
+		panic(err)
+	}
+
+	return ctx
+}
+
+func isZeroTrust(secOpts map[string]string) bool {
+	return secOpts != nil && secOpts["Mode"] == ZeroTrustKey
+}
+
+func HttpTransportFromService(secretProvider interfaces.SecretProviderExt, serviceInfo config.ServiceInfo, lc logger.LoggingClient) http.RoundTripper {
+	roundTripper := http.DefaultTransport
+	if isZeroTrust(serviceInfo.SecurityOptions) {
+		lc.Infof("zero trust client detected for service: %s", serviceInfo.Host)
+		roundTripper = createZitifiedTransport(secretProvider, serviceInfo.SecurityOptions[OpenZitiControllerKey], lc)
+	}
+	return roundTripper
+}
+
+func HttpTransportFromClient(secretProvider interfaces.SecretProviderExt, clientInfo *config.ClientInfo, lc logger.LoggingClient) http.RoundTripper {
+	roundTripper := http.DefaultTransport
+	if isZeroTrust(clientInfo.SecurityOptions) {
+		lc.Infof("zero trust client detected for client: %s", clientInfo.Host)
+		roundTripper = createZitifiedTransport(secretProvider, clientInfo.SecurityOptions[OpenZitiControllerKey], lc)
+	}
+	return roundTripper
+}
+
+func createZitifiedTransport(secretProvider interfaces.SecretProviderExt, ozController string, lc logger.LoggingClient) http.RoundTripper {
+	jwt, errJwt := secretProvider.GetSelfJWT()
+	if errJwt != nil {
+		lc.Errorf("could not load jwt: %v", errJwt)
+		return nil
+	}
+	ctx := AuthToOpenZiti(ozController, jwt)
+
+	zitiContexts := ziti.NewSdkCollection()
+	zitiContexts.Add(ctx)
+
+	zitiTransport := http.DefaultTransport.(*http.Transport).Clone() // copy default transport
+	zitiTransport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		dialer := zitiContexts.NewDialerWithFallback(ctx /*&net.Dialer{}*/, nil)
+		return dialer.Dial(network, addr)
+	}
+	return zitiTransport
+}
--- a/bootstrap/handlers/auth_middleware.go
+++ b/bootstrap/handlers/auth_middleware.go
@@ -22,12 +22,13 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/v3/clients/logger"
-
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/interfaces"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/secret"
+	"github.com/edgexfoundry/go-mod-core-contracts/v3/clients/logger"
 
 	"github.com/labstack/echo/v4"
+	"github.com/openziti/sdk-golang/ziti"
+	"github.com/openziti/sdk-golang/ziti/edge"
 )
 
 // VaultAuthenticationHandlerFunc prefixes an existing HandlerFunc
@@ -54,6 +55,24 @@ func VaultAuthenticationHandlerFunc(secretProvider interfaces.SecretProviderExt,
 			w := c.Response()
 			authHeader := r.Header.Get("Authorization")
 			lc.Debugf("Authorizing incoming call to '%s' via JWT (Authorization len=%d)", r.URL.Path, len(authHeader))
+
+			zitiCtx := r.Context().Value("zitiContext")
+			if zitiCtx != nil {
+				zc := *zitiCtx.(*ziti.Context)
+				zi, _ := zc.GetCurrentIdentity()
+
+				lc.Infof("Authorizing incoming connection via OpenZiti for %s", *zi.Name)
+				return inner(c)
+			}
+
+			zitiCtx = r.Context().Value("zero.trust.identityName")
+			if zitiCtx != nil {
+				zitiEdgeConn := zitiCtx.(edge.Conn)
+
+				lc.Infof("Authorizing incoming connection via OpenZiti for %s", zitiEdgeConn.SourceIdentifier())
+				return inner(c)
+			}
+
 			authParts := strings.Split(authHeader, " ")
 			if len(authParts) >= 2 && strings.EqualFold(authParts[0], "Bearer") {
 				token := authParts[1]
--- a/bootstrap/handlers/clients.go
+++ b/bootstrap/handlers/clients.go
@@ -58,12 +58,15 @@ func (cb *ClientsBootstrap) BootstrapHandler(
 	lc := container.LoggingClientFrom(dic.Get)
 	config := container.ConfigurationFrom(dic.Get)
 	cb.registry = container.RegistryFrom(dic.Get)
-	jwtSecretProvider := secret.NewJWTSecretProvider(container.SecretProviderExtFrom(dic.Get))
 
 	if config.GetBootstrap().Clients != nil {
 		for serviceKey, serviceInfo := range *config.GetBootstrap().Clients {
 			var url string
 			var err error
+
+			sp := container.SecretProviderExtFrom(dic.Get)
+			jwtSecretProvider := secret.NewJWTSecretProvider(sp)
+			sp.SetHttpTransport(container.HttpTransportFromClient(sp, serviceInfo, lc))
 
 			if !serviceInfo.UseMessageBus {
 				url, err = cb.getClientUrl(serviceKey, serviceInfo.Url(), startupTimer, dic, lc)
--- a/bootstrap/handlers/httpserver.go
+++ b/bootstrap/handlers/httpserver.go
@@ -20,9 +20,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	edge_apis "github.com/openziti/sdk-golang/edge-apis"
+	"github.com/openziti/sdk-golang/ziti"
+	"github.com/openziti/sdk-golang/ziti/edge"
+	"net"
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -43,6 +48,10 @@ type HttpServer struct {
 	router           *echo.Echo
 	isRunning        bool
 	doListenAndServe bool
+}
+
+type ZitiContext struct {
+	c *ziti.Context
 }
 
 // NewHttpServer is a factory method that returns an initialized HttpServer receiver struct.
@@ -124,6 +133,9 @@ func (b *HttpServer) BootstrapHandler(
 		Timeout: timeout,
 	}))
 
+	zc := &ZitiContext{}
+
+	b.router.Use(HandleOpenZiti(zc))
 	b.router.Use(RequestLimitMiddleware(bootstrapConfig.Service.MaxRequestSize, lc))
 
 	b.router.Use(ProcessCORS(bootstrapConfig.Service.CORSConfiguration))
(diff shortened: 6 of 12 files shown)


bootstrap/container/logging.go
```diff
@@ -32,3 +35,47 @@ func LoggingClientFrom(get di.Get) logger.LoggingClient {
...
+	// Create a new logger instance
+	hook := &LogrusAdaptor{
+		lc: l,
+	}
+	logrus.AddHook(hook)
+	logrus.SetFormatter(&logrus.TextFormatter{
+		DisableColors: true,
+	})
+	logrus.SetOutput(io.Discard)
+}
```
lenny-intel, line 81: I don't think this belongs in the `container` package. This package is for setting & getting objects from the DIC.

How about putting it in `utils` package?

go.mod
```diff
@@ -14,29 +14,53 @@ require (
 	github.com/labstack/echo/v4 v4.11.4
 	github.com/mitchellh/copystructure v1.2.0
 	github.com/mitchellh/mapstructure v1.5.0
+	github.com/openziti/sdk-golang v0.22.21
 	github.com/rcrowley/go-metrics v0.0.0-20201227073835-cf1acfcdf475
+	github.com/sirupsen/logrus v1.9.3
```
lenny-intel, line 19: These two new dependencies are new to EdgeX and need to go thru the vetting process and TSC approval.
See process here: https://wiki.edgexfoundry.org/display/FA/Vetting+Process+for+3rd+Party+Dependencies

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +168,70 @@ func (b *HttpServer) BootstrapHandler(
...
+		var ln net.Listener
+		switch bootstrapConfig.Service.SecurityOptions["Mode"] {
+		case "zerotrust":
+			secretProvider := container.SecretProviderExtFrom(dic.Get)
+			var zitiCtx ziti.Context
+			var ctxErr error
+			if secretProvider != nil {
+				jwt, jwtErr := secretProvider.GetSelfJWT()
+				if jwtErr != nil {
+					lc.Errorf("could not load jwt: %v", jwtErr)
```
lenny-intel, line 181: Should return log error and return false.

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +168,70 @@ func (b *HttpServer) BootstrapHandler(
...
+				}
+				lc.Info("using zerotrust - look at you go: " + jwt)
```
lenny-intel, line 183: Should we be logging the JWT??

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +168,70 @@ func (b *HttpServer) BootstrapHandler(
...
 		b.isRunning = true
-		err := server.ListenAndServe()
+
+		var ln net.Listener
+		switch bootstrapConfig.Service.SecurityOptions["Mode"] {
+		case "zerotrust":
+			secretProvider := container.SecretProviderExtFrom(dic.Get)
+			var zitiCtx ziti.Context
+			var ctxErr error
+			if secretProvider != nil {
```
lenny-intel, line 178: secretProvider  should never be nil. It will either be the `SecureProvider` or the `InsecureProvider`

If it is nil, then log error and return false.

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +168,70 @@ func (b *HttpServer) BootstrapHandler(
...
+					panic("here?")
+				} else {
+					zitiCtx, ctxErr = ziti.LoadContext(ozIdFile)
+					if ctxErr != nil {
+						panic(ctxErr)
+					}
+				}
+
+				ziti.DefaultCollection.Add(zitiCtx)
+			}
```
lenny-intel, line 215: Should never get here. So code not needed. See previous comment

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +168,70 @@ func (b *HttpServer) BootstrapHandler(
...
+
+			zc.c = &zitiCtx
+		case "http":
+		default:
+			lc.Warnf("using ListenMode 'http' at %s", addr)
+			ln, err = net.Listen("tcp", addr)
+		}
+		server.ConnContext = mutator
+
+		if err == nil {
```
lenny-intel, line 231: Check not needed if properly log err and return `false` above.

lenny-intel (changes requested):

bootstrap/handlers/httpserver.go
dovholuknf: definitely not.  :) i had left it here to aid with debugging. i planned to take it out after things were working. since they're working now, i'll remove it

bootstrap/container/zerotrust.go
```diff
@@ -0,0 +1,89 @@
...
+const (
+	OpenZitiControllerKey = "OpenZitiController"
+	ZeroTrustKey          = "zerotrust"
+)
+
+func AuthToOpenZiti(ozController, jwt string) ziti.Context {
+	openZitiRootUrl := "https://" + ozController
+	if !strings.Contains(openZitiRootUrl, "://") {
+		openZitiRootUrl = "https://" + ozController
+	}
```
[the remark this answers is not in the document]
dovholuknf: ha yes. thanks, I reduced it further by simply reusing the string that's passed in.

bootstrap/container/zerotrust.go
```diff
@@ -0,0 +1,89 @@
...
+
+	cfg := &ziti.Config{
+		ZtAPI:       openZitiRootUrl + "/edge/client/v1",
+		Credentials: credentials,
+	}
+	cfg.ConfigTypes = append(cfg.ConfigTypes, "all")
+
+	ctx, ctxErr := ziti.NewContext(cfg)
+	if ctxErr != nil {
+		panic(ctxErr)
```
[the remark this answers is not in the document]
dovholuknf: i wasn't sure how to handle this. if you've asked to use security and you cannot authenticate, I opted to have the whole process panic since "nothing" will work. but i'm happy to just return an error... i'll update it and we can discuss further

bootstrap/handlers/httpserver.go
dovholuknf: instead, i'll set `err` and let it continue down to line 231 where it checks the err and then either runs the server or exits in the next if block... that seems to be the pattern here. (note that this is within a goroutine and doesn't follow the established pattern of returning false)

bootstrap/handlers/httpserver.go
dovholuknf: moved the server.serve to within both if's instead and adopted the existing pattern of not returning false, but instead setting an error to get through to the `if err != nil && err != http.ErrServerClosed {` block

bootstrap/handlers/auth_middleware.go
```diff
@@ -54,6 +55,24 @@ func VaultAuthenticationHandlerFunc(secretProvider interfaces.SecretProviderExt,
 			w := c.Response()
 			authHeader := r.Header.Get("Authorization")
 			lc.Debugf("Authorizing incoming call to '%s' via JWT (Authorization len=%d)", r.URL.Path, len(authHeader))
+
+			zitiCtx := r.Context().Value("zitiContext")
+			if zitiCtx != nil {
+				zc := *zitiCtx.(*ziti.Context)
+				zi, _ := zc.GetCurrentIdentity()
+
+				lc.Infof("Authorizing incoming connection via OpenZiti for %s", *zi.Name)
```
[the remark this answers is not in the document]
dovholuknf: sure!

bootstrap/handlers/auth_middleware.go
```diff
@@ -54,6 +55,24 @@ func VaultAuthenticationHandlerFunc(secretProvider interfaces.SecretProviderExt,
 			w := c.Response()
 			authHeader := r.Header.Get("Authorization")
 			lc.Debugf("Authorizing incoming call to '%s' via JWT (Authorization len=%d)", r.URL.Path, len(authHeader))
+
+			zitiCtx := r.Context().Value("zitiContext")
+			if zitiCtx != nil {
+				zc := *zitiCtx.(*ziti.Context)
+				zi, _ := zc.GetCurrentIdentity()
```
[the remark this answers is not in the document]
dovholuknf: no good reason tbh. probably just missed cleaning this up. thx

bootstrap/container/zerotrust.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func isZeroTrust(secOpts map[string]string) bool {
+	return secOpts != nil && secOpts["Mode"] == ZeroTrustKey
+}
+
+func HttpTransportFromService(secretProvider interfaces.SecretProviderExt, serviceInfo config.ServiceInfo, lc logger.LoggingClient) http.RoundTripper {
+	roundTripper := http.DefaultTransport
+	if isZeroTrust(serviceInfo.SecurityOptions) {
+		lc.Infof("zero trust client detected for service: %s", serviceInfo.Host)
```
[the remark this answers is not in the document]
dovholuknf: sure. changed to debug

bootstrap/container/logging.go
dovholuknf: this file/package/function was already here. 
```
func LoggingClientFrom(get di.Get) logger.LoggingClient {
```
I placed these here for that reason.  I'm happy to move the additions or the whole file if you want? I just put these "near" the same location as that function.

lenny-intel: LoggingClientFrom should and existing code should stay. Your should move since that are not dic container related. i.e. they are beyond the scope of `container`

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +168,70 @@ func (b *HttpServer) BootstrapHandler(
...
+			serviceName := bootstrapConfig.Service.SecurityOptions["OpenZitiServiceName"]
+			ln, err = zitiCtx.Listen(serviceName)
+			if err != nil {
+				panic("could not bind service " + serviceName + ": " + err.Error())
+			}
+
+			zc.c = &zitiCtx
+		case "http":
+		default:
+			lc.Warnf("using ListenMode 'http' at %s", addr)
```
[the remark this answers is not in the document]
lenny-intel: This one is still a Warnf. PLease change to Inof

bootstrap/utils/logging.go
```diff
@@ -0,0 +1,60 @@
+package utils
```
lenny-intel, line 1: Ahhhg! This causes a cyclic import. :-(

Please create a logging package to avoid this and name the file logging/logrus-adaptor.go

dovholuknf: yeah... i didn't spot it at first. I'll work through it.

bootstrap/handlers/clients.go
```diff
@@ -58,13 +59,23 @@ func (cb *ClientsBootstrap) BootstrapHandler(
...
 			var url string
 			var err error
 
+			sp := container.SecretProviderExtFrom(dic.Get)
+			jwtSecretProvider := secret.NewJWTSecretProvider(sp)
+			if rt, transpErr := zerotrust.HttpTransportFromClient(sp, serviceInfo, lc); transpErr != nil {
+				lc.Errorf("could not obtain an http client for use with zero trust provider: %v", transpErr)
+				return false
+			} else {
+				if sp != nil {
```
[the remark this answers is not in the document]
dovholuknf: I had it that way but the test failed again. It must be a different test had failed and I didn't notice it. I'll take a closer look. With the nil check the suite passed so I didn't dig much further. I'll look

bootstrap/handlers/auth_middleware.go
```diff
@@ -54,6 +56,27 @@ func VaultAuthenticationHandlerFunc(secretProvider interfaces.SecretProviderExt,
 			w := c.Response()
 			authHeader := r.Header.Get("Authorization")
 			lc.Debugf("Authorizing incoming call to '%s' via JWT (Authorization len=%d)", r.URL.Path, len(authHeader))
+
+			zitiCtx := r.Context().Value("zitiContext")
```
lenny-intel, line 60: We should be doing this only when zerotrust is enabled in configuration.

lenny-intel (changes requested):

bootstrap/handlers/clients.go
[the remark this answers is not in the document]
dovholuknf: Thanks for the offline help. The issue was that the type assertion would fail as I'd added two new functions to the interface. Reran mockery and fixed the assertion and was able to remove the if check

bootstrap/handlers/auth_middleware.go
dovholuknf: this is older and I think is no longer used in favor of the other if block below this one. i need to build all the pieces and test it out along with the `HandleOpenZiti` function that has no apparent use. I think this is a holdover from the original implementation and I think it needs a bit of refactoring.

As to "We should be doing this only when zerotrust is enabled in configuration.". Do you want another function on `SecretProviderExt` such as `IsZeroTrust` ?

dovholuknf: I added two more interface functions for your consideration. see what you think.
```

	// ZeroTrustEnabled returns whether zero trust principles are enabled
	ZeroTrustEnabled() bool

	// EnableZeroTrust marks the provider as being zero trust enabled
	EnableZeroTrust()
```

bootstrap/container/secret.go
```diff
@@ -40,7 +40,8 @@ var SecretProviderExtName = di.TypeInstanceToName((*interfaces.SecretProvider)(n
 // SecretProviderExtFrom helper function queries the DIC and returns the interfaces.SecretProviderExt
 // implementation.
 func SecretProviderExtFrom(get di.Get) interfaces.SecretProviderExt {
-	provider, ok := get(SecretProviderExtName).(interfaces.SecretProviderExt)
+	sp := get(SecretProviderExtName)
+	provider, ok := sp.(interfaces.SecretProviderExt)
```
lenny-intel: This change is not needed. Left over from debugging?

bootstrap/bootstrap.go
```diff
@@ -251,3 +255,16 @@ func registerMetrics(metricsManager interfaces.MetricsManager, metrics map[strin
...
+	// Create a new logger instance
+	hook := &utils.LogrusAdaptor{
+		Logger: l,
+	}
+	logrus.AddHook(hook)
+	logrus.SetFormatter(&logrus.TextFormatter{
+		DisableColors: true,
+	})
+	logrus.SetOutput(io.Discard)
+}
```
lenny-intel, line 270: I liked it better when this was in the utils/logging.go with the others. Only change needed is to pass  LoggingClient instead of dic which get rid of the cyclic import.
```suggestion
func adaptLogrusBasedLogging(lc LoggingClientFrom) {
	// Create a new logger instance
	hook := &utils.LogrusAdaptor{
		Logger: lc,
	}
	logrus.AddHook(hook)
	logrus.SetFormatter(&logrus.TextFormatter{
		DisableColors: true,
	})
	logrus.SetOutput(io.Discard)
}
```

bootstrap/handlers/httpserver.go
```diff
@@ -204,3 +280,19 @@ func RequestLimitMiddleware(sizeLimit int64, lc logger.LoggingClient) echo.Middl
...
 }
+
+func mutator(srcCtx context.Context, c net.Conn) context.Context {
+	if zitiConn, ok := c.(edge.Conn); ok {
+		return context.WithValue(srcCtx, "zero.trust.identityName", zitiConn)
+	}
+	return srcCtx
+}
+
+func HandleOpenZiti(zitiCtx *ZitiContext) echo.MiddlewareFunc {
```
lenny-intel, line 291: `zitiCtx` isn't used. Am I missing something here?

bootstrap/secret/secure.go
```diff
@@ -477,3 +480,19 @@ func (p *SecureProvider) GetSelfJWT() (string, error) {
...
+
+func (p *SecureProvider) SetHttpTransport(rt http.RoundTripper) {
+	p.httpRoundTripper = rt
+}
+
+func (p *SecureProvider) IsZeroTrustEnabled() bool {
+	return p.zeroTrustEnabled
+}
+
+func (p *SecureProvider) ZeroTrustEnabled() {
```
lenny-intel, line 496: Match how define in the interface.

```suggestion
func (p *SecureProvider) EnableZeroTrust() {
```

bootstrap/secret/insecure.go
```diff
@@ -245,3 +246,19 @@ func (p *InsecureProvider) GetSelfJWT() (string, error) {
...
+
+func (p *InsecureProvider) SetHttpTransport(_ http.RoundTripper) {
+	//empty on purpose
+}
+
+func (p *InsecureProvider) IsZeroTrustEnabled() bool {
+	return false
+}
+
+func (p *InsecureProvider) ZeroTrustEnabled() {
```
lenny-intel, line 262: Match how define in the interface.

```suggestion
func (p *InsecureProvider) EnableZeroTrust() {
```

bootstrap/interfaces/secret.go
```diff
@@ -68,4 +71,16 @@ type SecretProviderExt interface {
...
 	IsJWTValid(jwt string) (bool, error)
+
+	// HttpTransport returns the http.RoundTripper to be used by http-based clients
+	HttpTransport() http.RoundTripper
+
+	// SetHttpTransport sets the http.RoundTripper to be used by http-based clients
+	SetHttpTransport(rt http.RoundTripper)
+
+	// ZeroTrustEnabled returns whether zero trust principles are enabled
+	ZeroTrustEnabled() bool
```
lenny-intel, line 82: Match name used in the implementations
```suggestion
	IsZeroTrustEnabled() bool
```

lenny-intel (changes requested):

bootstrap/bootstrap.go
dovholuknf: i like that idea. much cleaner for sure thx

bootstrap/container/secret.go
dovholuknf: i'd noticed that and fixed in the last commit . thx

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +169,70 @@ func (b *HttpServer) BootstrapHandler(
...
-		err := server.ListenAndServe()
+
+		switch bootstrapConfig.Service.SecurityOptions["Mode"] {
+		case "zerotrust":
+			secretProvider := container.SecretProviderExtFrom(dic.Get)
+			if secretProvider == nil {
+				err = errors.New("secret provider is nil. cannot proceed with zero trust configuration")
+				break
+			}
+			secretProvider.IsZeroTrustEnabled() //mark the secret provider as zero trust enabled
```
lenny-intel, line 180:
```suggestion
			secretProvider.EnableZeroTrust() //mark the secret provider as zero trust enabled
```

lenny-intel (changes requested):

bootstrap/secret/secure.go
lenny-intel: This changes wasn't made

bootstrap/secret/insecure.go
lenny-intel: This change wasn't made

lenny-intel (approved):
LGTM, Nice work!

go.mod
dovholuknf: I've updated the PR but i dunno how to encourage the -dependency and -security-review labels to be added to the PR.

lenny-intel: Added

lenny-intel:
@jumpingliu , @cloudxxx8 , @JamesKButcher , there are two new dependencies added by this PR. @clint has done the vetting and put the information in the PR description per our process. Though it is not explicit in the documented process, we have always had a TSC vote to add new 3rd party dependencies,

bootstrap/secret/secure.go
```diff
@@ -65,6 +66,8 @@ type SecureProvider struct {
 	securityConsulTokenDuration        gometrics.Timer
 	securityRuntimeSecretTokenDuration gometrics.Timer
 	securityGetSecretDuration          gometrics.Timer
+	httpRoundTripper                   http.RoundTripper
+	zeroTrustEnabled                   bool
```
cloudxxx8, line 70: Since there are some functions to write those variable, the mutex should be added to protect these variable access to prevent the race condition

cloudxxx8 (changes requested):

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +169,70 @@ func (b *HttpServer) BootstrapHandler(
...
+				err = fmt.Errorf("could not bind service " + serviceName + ": " + listenErr.Error())
+				break
+			}
+
+			zc.c = &zitiCtx
+			lc.Infof("using ListenMode 'zerotrust' at %s", addr)
+			err = server.Serve(ln)
+		case "http":
+		default:
+			lc.Infof("using ListenMode 'http' at %s", addr)
```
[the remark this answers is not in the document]
dovholuknf: yes thanks for catching that. years of java -- get me all the time in go. :) will fix in next commit by adding fallthrough

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +169,70 @@ func (b *HttpServer) BootstrapHandler(
 		}()
 
 		b.isRunning = true
-		err := server.ListenAndServe()
+
+		switch bootstrapConfig.Service.SecurityOptions["Mode"] {
```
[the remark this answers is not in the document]
dovholuknf: i'll tolower in the next commit and move "Mode" to a constant in this file as suggested

dovholuknf:
i've amended the initial commit and signed it to fix the DCO and addressed the PR comments.

bootstrap/secret/secure.go
dovholuknf: see commit https://github.com/edgexfoundry/go-mod-bootstrap/pull/659/commits/d4cd1881d7d5a3bda485488f9f663197170b8a05

bootstrap/handlers/httpserver.go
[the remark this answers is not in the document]
dovholuknf: see commit https://github.com/edgexfoundry/go-mod-bootstrap/pull/659/commits/d4cd1881d7d5a3bda485488f9f663197170b8a05

bootstrap/handlers/httpserver.go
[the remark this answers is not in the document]
dovholuknf: see commit https://github.com/edgexfoundry/go-mod-bootstrap/pull/659/commits/d4cd1881d7d5a3bda485488f9f663197170b8a05

cloudxxx8:
DCO may be failed by one of your commits, so you can squash all into one commit and make sure it can pass DCT

go.mod
cloudxxx8, line 1: since go-mod-core-contracts has merged the related PR, please upgrade its version as well

dovholuknf: thanks. updated to the latest -dev.13

dovholuknf:
thanks. fixed. i really do need to setup 'auto signoff'...

cloudxxx8 (approved):
LGTM

bootstrap/handlers/httpserver.go
```diff
@@ -156,7 +172,70 @@ func (b *HttpServer) BootstrapHandler(
...
+			}
+
+			serviceName := bootstrapConfig.Service.SecurityOptions[config.OpenZitiServiceNameKey]
+			ln, listenErr := zitiCtx.Listen(serviceName)
+			if listenErr != nil {
+				err = fmt.Errorf("could not bind service " + serviceName + ": " + listenErr.Error())
+				break
+			}
+
+			zc.c = &zitiCtx
```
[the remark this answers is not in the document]
dovholuknf: from what I see, it appears to be totally vestigal. I don't find any references to it whatsoever other than this one line. I expect at one point a function was defined on the struct, the function removed, leaving this behind.

PR #678, opened 2024-03-18
feat: allow clients to be zero trust

dovholuknf:
following https://github.com/edgexfoundry/go-mod-bootstrap/issues/650 and https://github.com/edgexfoundry/go-mod-bootstrap/pull/659, when enabling app-services to be able to participate with the zero trust overlay it's necessary to mark the secret provider if the key zero trust mode is set

## PR Checklist

  <link to docs PR>

### DIFF
--- a/bootstrap/handlers/clients.go
+++ b/bootstrap/handlers/clients.go
@@ -18,7 +18,6 @@ package handlers
 import (
 	"context"
 	"fmt"
-	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/config"
 	"sync"
 	"time"
 
@@ -30,6 +29,7 @@ import (
 	"github.com/edgexfoundry/go-mod-registry/v3/pkg/types"
 	"github.com/edgexfoundry/go-mod-registry/v3/registry"
 
+	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/config"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/secret"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/startup"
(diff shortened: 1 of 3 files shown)


bootstrap/handlers/clients.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/edgexfoundry/go-mod-registry/v3/pkg/types"
 	"github.com/edgexfoundry/go-mod-registry/v3/registry"
 
+	bscfg "github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/config"
```
lenny-intel, line 32: Alias not needed
```suggestion
	"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/config"
```

bootstrap/handlers/clients.go
```diff
@@ -67,6 +68,9 @@ func (cb *ClientsBootstrap) BootstrapHandler(
 
 			sp := container.SecretProviderExtFrom(dic.Get)
 			jwtSecretProvider := secret.NewJWTSecretProvider(sp)
+			if serviceInfo.SecurityOptions[bscfg.SecurityModeKey] == zerotrust.ConfigKey {
```
lenny-intel, line 71: Naming of value constant should be `zerotrust.ZeroTrustMode`. i.e it isn't a key...
```suggestion
			if serviceInfo.SecurityOptions[bscfg.SecurityModeKey] == zerotrust.ZeroTrustMode{
```

lenny-intel (changes requested):

bootstrap/handlers/clients.go
dovholuknf: I like the idea, how about SecurityMode instead since it 'SecurityModeKey'?

dovholuknf: actually - neermind. ZeroTrustMode makes more sense ;)

bootstrap/handlers/clients.go
dovholuknf: fine with me, just meant i had to refactor the existing 'config' variable. done

lenny-intel (approved):
LGTM

PR #749, opened 2024-08-20
feat: Add new env to support -o flag

### LINKED ISSUE #747: Add an option to make the environment variable config override take priority
# 🚀 Feature Request

 This feature request is for all

In EdgeX v3, if the config provider is enabled, environment variable override only takes effect and push to the config provider in the first start up.  However, some users still use to set the environment variable override and use them to control the system.

Create a new EDGEX_OVERWRITE_CONFIG environment variable, like others in https://docs.edgexfoundry.org/3.1/microservices/configuration/CommonEnvironmentVariables/

The effect will be the same as -o/--overwrite flag
https://docs.edgexfoundry.org/3.1/microservices/configuration/CommonCommandLineOptions/#overwrite

If it is set to "true",overwrite configuration in provider with local configuration and environment variables.
The default value is "false", and the empty string and all other values treat as "false"

cloudxxx8:
Since the flag is hard to use for the container environment, create a new env EDGEX_OVERWRITE_CONFIG to provide the same feature support as -o/--overwrite flag.

Close #747

## PR Checklist

  <link to docs PR>

## Testing Instructions

1. Start the EdgeX with the default configs and config provider.
2. Modify the docker-compose file with the environment variable `EDGEX_OVERWRITE_CONFIG: true` and set other environment variable to change the private or common config.
3. Check the config provider if the new change of the environment variable overwrite is pushed.

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -75,7 +75,6 @@ type Processor struct {
 	wg                 *sync.WaitGroup
 	configUpdated      UpdatedStream
 	dic                *di.Container
-	overwriteConfig    bool
 	providerHasConfig  bool
 	commonConfigClient configuration.Client
 	appConfigClient    configuration.Client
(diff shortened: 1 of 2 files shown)


jinlinGuan (approved):
LGTM

PR #894, opened 2025-08-15
feat: Ensure new configurations from the file are applied

weichou1229:
After upgrading EdgeX, new configurations may be introduced. Ensure new configurations from the YAML file are pushed to the core-keeper.

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact
- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

- deploy old version (such tag [v4.1.0-dev.46](https://github.com/edgexfoundry/edgex-go/releases/tag/v4.1.0-dev.46)) edgex-go
- shut down and deploy the latest edgex-go with this PR.
- verify the new configs are pushed to core-keeper

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -278,6 +278,13 @@ func (cp *Processor) loadConfigByProvider(
 	}
 
 	if cp.providerHasConfig && !cp.getOverwriteConfig() {
+		// After upgrading EdgeX, new configurations may be introduced.
+		// Ensure new configurations from the YAML file are pushed to the core-keeper.
+		err = cp.putMissingPrivateConfigToConfigProvider()
+		if err != nil {
+			return err
+		}
+
 		privateServiceConfig, err := copyConfigurationStruct(serviceConfig)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


bootstrap/config/config.go
```diff
@@ -1271,3 +1280,76 @@ func GetInsecureSecretDataFullPath(secretName, key string) string {
...
+		if value == nil {
+			continue
+		}
+		hasSubConfig, err := configClient.HasSubConfiguration(key)
+		if err != nil {
+			return fmt.Errorf("failed to check config %s: %s", key, err.Error())
+		}
+		if hasSubConfig {
+			continue
+		}
```
judehung, line 1300: Add a comment to indicate the rationale why skip the key when it has subConfig

bootstrap/config/config.go
```diff
@@ -1271,3 +1280,76 @@ func GetInsecureSecretDataFullPath(secretName, key string) string {
...
+		if err = configClient.PutConfigurationValue(key, []byte(fmt.Sprint(value))); err != nil {
+			return fmt.Errorf("failed to push common configuration key %s with value %v: %s", key, value, err.Error())
```
judehung, line 1302: From the error log, does the code here only handle common config? How about the private config?

bootstrap/config/config.go
```diff
@@ -1271,3 +1280,76 @@ func GetInsecureSecretDataFullPath(secretName, key string) string {
...
+			key = origKey
+			continue
+		}
+
+		kv = buildKeyValues(vdata, kv, key)
+		key = origKey
+	}
+
+	return kv
+}
```
judehung, line 1355: The current function comment doesn’t clearly convey the purpose of this helper, making it hard to judge whether its recursive logic is correct.

From the code, I can only guess that the function’s intent is to flatten a nested map into a single-level map? If that’s correct, please update the comment to explicitly describe this purpose. Also, since this is a utility function, consider to move this func to bootstrap/utils/utils.go and this func should have unit tests to verify that its behavior matches the intended goal.

Issue #911, opened 2025-10-15
Insecure Secret Provider fails with secret names containing '/'


FelixTing:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: EdgeX services that provide the POST /secrets API 

### Is this a regression?

 no

### Description and Minimal Reproduction [**REQUIRED**]
Both `GetSecret` and `StoreSecret` in [bootstrap/secret/insecure.go](https://github.com/edgexfoundry/go-mod-bootstrap/blob/main/bootstrap/secret/insecure.go) currently use secretName directly. This causes problems when secretName contains special characters (especially "/"), which CoreKeeper uses as a key delimiter. As a result, EdgeX services would fail to retrieve secrets correctly when such names are used.

## 🌍  Your Environment
**EdgeX Version [**REQUIRED**]:** main

PR #913, opened 2025-10-16
fix!: URL encode secret names in InsecureProvider

FelixTing:
BREAKING CHANGE: Secret names with special characters (such as `/`) will be URL encoded. Existing secrets containing special characters must be recreated.

fix: #911 

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
  <link to docs PR>

### DIFF
--- a/bootstrap/secret/insecure.go
+++ b/bootstrap/secret/insecure.go
@@ -1,5 +1,6 @@
 /*******************************************************************************
  * Copyright 2020-2023 Intel Corporation
+ * Copyright 2025 IOTech Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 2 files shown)


bootstrap/secret/insecure.go
```diff
@@ -119,15 +123,17 @@ func (p *InsecureProvider) StoreSecret(secretName string, secrets map[string]str
 		return errors.NewCommonEdgeX(errors.KindNotAllowed, "can't store secrets. ConfigurationProvider is not in use or has not been properly initialized", nil)
 	}
 
+	encodedSecretName := url.QueryEscape(secretName)
```
judehung: Same comment as above.

